PR #21, opened 2026-07-25
perf(metal): skip redundant fp32 cast in turboquant_scalar_quantize for fp16 input

### LINKED ISSUE #20: perf(metal): turboquant_scalar_quantize forces redundant fp32 cast on fp16 input
## Summary

`turboquant_scalar_quantize` (the nearest-centroid scalar-quantize Metal kernel used by `TurboQuantMSE`, `TurboQuantRVQ`, and the Hadamard preconditioner) unconditionally casts its input tensor to fp32 before dispatching the Metal kernel — even when the input is already fp16, which is the dtype used on every real KV-cache quantize call in this codebase.

This was found during a static audit of all 14 Metal kernels in `veloxquant_mlx/metal/` (bit-packing, scalar quant/dequant, RaBitQ encode/attend/prefill, RVQ attend, QJL, CommVQ, VecInfer, fused SDPA). Full audit notes below; this issue tracks just the one fastest/cheapest fix from that audit.

## Where

`veloxquant_mlx/metal/_scalar_quant.py`, function `turboquant_scalar_quantize`, line 187 (pre-fix):

```python
*leading, d = x.shape
flat_x    = x.reshape(-1).astype(mx.float32)   # <-- unconditional cast
cents_f32 = centroids.astype(mx.float32)
N         = flat_x.size

outputs = _scalar_quantize_kernel(b)(
    inputs=[flat_x, cents_f32],
    ...
```

## Why this is wasteful

The Metal kernel body (`_SCALAR_QUANTIZE_SRC`) already does the fp32 promotion **inside** the kernel, per-element, in-register:

```metal
uint  elem = thread_position_in_grid.x;
float val  = float(x[elem]);   // <-- already upcasts on read
```

So the Python-side `.astype(mx.float32)` on line 187 does nothing useful when `x` is fp16 — it just forces MLX to run a **second, separate dispatch** (a cast kernel over all `N` elements) purely to convert dtype, before the actual quantize kernel even launches. Every `update_and_fetch` call on a fp16 KV-cache pays this extra kernel-launch + memory-read/write pass for no numerical benefit, since the Metal kernel would produce the identical result reading `half` directly.

This pattern is already handled correctly for a sibling kernel — `veloxquant_mlx/metal/_vecinfer.py`'s `_dequant_kernel`/`_quantize_kernel` factories (lines 263–286) key their pipeline-state cache on `str(dtype)` and compile a separate kernel variant per input dtype instead of forcing a cast. `_scalar_quant.py` should follow the same pattern.

## What change is needed

1. Template `_scalar_quantize_kernel(b, dtype)` so the MLX-compiled pipeline is cached per `(b, dtype)` pair, not just `b` — mirroring `_vecinfer.py`'s existing dtype-keyed cache.
2. In `turboquant_scalar_quantize`, only cast to fp32 when the input dtype isn't already one of Metal's native float types (`mx.float16`, `mx.float32`); pass fp16 straight through otherwise.
3. Leave `centroids` cast to fp32 unchanged — it's a tiny `[2^b]` array (≤16 elements) and always fp32 by contract, so casting it costs nothing and isn't the hot path.
4. No change to kernel semantics, output dtype (`uint8` indices), or numerical behavior — this is purely removing a redundant dispatch, not an algorithm change.

## Validation plan

- All existing correctness tests in `veloxquant_mlx/tests/metal/test_turboquant_kernels.py` (`test_scalar_quantize_correctness` × 4 bit-widths, `test_scalar_dequantize_correctness` × 4, `test_quantize_dequantize_roundtrip`, plus unrelated `test_bit_pack_roundtrip` × 9 as a regression check) must still pass unchanged.
- New benchmark script isolating just the cast-vs-no-cast delta on fp16 input, at realistic KV-cache shapes (batch×head_dim from 64×128 up to 16384×128/256, matching this repo's own Falcon3/Gemma test shapes), before and after.
- Manual dtype-path check: confirm both fp16 and fp32 inputs still produce correct `uint8` output post-change (int/other dtypes still take the explicit-cast fallback path).

## Measured result (already validated on this branch, see linked PR)

Median of 100 iterations per shape, fp16 input, Apple Silicon GPU:

| shape (B×d) | b | before (ms) | after (ms) | speedup |
|---|---|---|---|---|
| 64×128 | 2 | 0.2455 | 0.2104 | 1.17× |
| 64×128 | 4 | 0.1839 | 0.1754 | 1.05× |
| 512×128 | 2 | 0.1848 | 0.1690 | 1.09× |
| 512×128 | 4 | 0.1841 | 0.1688 | 1.09× |
| 4096×128 | 2 | 0.2347 | 0.1922 | 1.22× |
| 4096×128 | 4 | 0.2493 | 0.2230 | 1.12× |
| 4096×256 | 2 | 0.3141 | 0.2213 | **1.42×** |
| 4096×256 | 4 | 0.3216 | 0.2374 | **1.35×** |
| 16384×128 | 2 | 0.4716 | 0.2661 | **1.77×** |
| 16384×128 | 4 | 0.4659 | 0.2863 | **1.63×** |

Gain grows with tensor size, as expected for removing a fixed-relative-cost O(N) dispatch. Up to **1.77×** at long-context shapes. All 20 existing correctness tests pass unchanged.

## Related findings (out of scope for this issue, tracked separately if pursued)

From the same audit pass, not addressed here:
- `_bit_packing.py` pack/unpack kernels lack bounds guards other kernels in the set already have (latent OOB risk when N isn't a multiple of the threadgroup size).
- `_vecinfer.py`'s fused key-path kernel (`_ENCODE_DECODE_FULL_SRC`) implements its Walsh-Hadamard Transform as an O(D²) dense matvec instead of the O(D log D) butterfly already correct in two sibling kernels (`_rabitq_encode.py`, `_scalar_quant.py`'s own Hadamard-quantize kernel) — largest single arithmetic-efficiency gap found, ~32× theoretical FMA reduction at D=256, not yet attempted.
- `fused_sdpa.py`'s fully-built fused SDPA kernel is confirmed dead code in production (never routed to from the real generation loop per its own inline comment).
- `_rvq_attend.py`'s fused decode+attend kernel caps its simdgroup count at `min(D, 32)`, leaving kv-axis parallelism on the table for head_dim > 32, unlike the proven 8-way flash-decoding split already in `_rabitq_attend.py`.

rajveer43:
## Summary

Fixes #20 — `turboquant_scalar_quantize` was forcing a redundant fp32 cast on fp16 input before dispatching its Metal kernel, even though the kernel already upcasts per-element in-register. This adds an extra MLX cast-kernel dispatch on every call for no numerical benefit.

- Templates the Metal pipeline cache on `(b, dtype)` instead of just `b`, mirroring the dtype-keyed cache pattern already used in `_vecinfer.py`'s `_dequant_kernel`/`_quantize_kernel`.
- Passes fp16/fp32 input straight through to the kernel; only non-native-float dtypes still take the explicit `.astype(mx.float32)` fallback.
- `centroids` cast is unchanged (tiny `[2^b]` array, always fp32 by contract, not the hot path).
- No change to kernel semantics, output dtype, or numerical behavior.

This branch also carries 3 prior, unrelated landing-page commits (Playground page, UI polish, Compression Lab RAM sync) that were already on `feat/metal-kernels` before this fix — see individual commits below.

## Measured results

Median of 100 iterations per shape, fp16 input, Apple Silicon GPU (`scripts/bench_scalar_quantize_dtype_cast.py`, included in this PR):

| shape (B×d) | b | before (ms) | after (ms) | speedup |
|---|---|---|---|---|
| 64×128 | 2 | 0.2455 | 0.2104 | 1.17× |
| 64×128 | 4 | 0.1839 | 0.1754 | 1.05× |
| 512×128 | 2 | 0.1848 | 0.1690 | 1.09× |
| 512×128 | 4 | 0.1841 | 0.1688 | 1.09× |
| 4096×128 | 2 | 0.2347 | 0.1922 | 1.22× |
| 4096×128 | 4 | 0.2493 | 0.2230 | 1.12× |
| 4096×256 | 2 | 0.3141 | 0.2213 | **1.42×** |
| 4096×256 | 4 | 0.3216 | 0.2374 | **1.35×** |
| 16384×128 | 2 | 0.4716 | 0.2661 | **1.77×** |
| 16384×128 | 4 | 0.4659 | 0.2863 | **1.63×** |

Gain grows with tensor size, consistent with removing a fixed-relative-cost O(N) dispatch.

## Test plan

- [x] `PYTHONPATH=. python -m pytest veloxquant_mlx/tests/metal/test_turboquant_kernels.py -k "scalar_quantize or scalar_dequantize or roundtrip"` — 20/20 passed, no regressions (including unrelated `test_bit_pack_roundtrip` cases as a sanity check that nothing else broke).
- [x] Manually verified both `mx.float16` and `mx.float32` inputs still produce correct `uint8` output.
- [x] Benchmark script run before and after the change; results captured above.


Som23Git (commented):
@rajveer43 Holding off on merge here.

1. **Merge conflicts with `master`** (`mergeable: CONFLICTING`).
2. **Scope mix:** the Metal cast fix in `_scalar_quant.py` is small and valuable (#20), but this PR also carries a large landing/playground diff. Please split — keep this PR (or a fresh one) to the Metal change + bench script only, and land landing work via the landing PR path.

Once rebased/split with green CI, the scalar-quantize cast removal looks good to merge on its own (measured speedups, existing tests as the gate).

Som23Git (commented):
@rajveer43 Holding off on merge here.

1. **Merge conflicts with `master`** (`mergeable: CONFLICTING`).
2. **Scope mix:** the Metal cast fix in `_scalar_quant.py` is small and valuable (#20), but this PR also carries a large landing/playground diff. Please split — keep this PR (or a fresh one) to the Metal change + bench script only, and land landing work via the landing PR path.

Once rebased/split with green CI, the scalar-quantize cast removal looks good to merge on its own (measured speedups, existing tests as the gate).

PR #30, opened 2026-07-28
fix(a2ats): correct four paper-fidelity deviations (#29)

### LINKED ISSUE #29: A2ATS-adapted: windowed RoPE rotates far keys (paper leaves them unrotated), b conflated with w, per-step distance gating frozen at write time
## Summary

Reviewed the `a2ats` implementation against the source paper ([A2ATS, ACL 2025 Findings](https://aclanthology.org/2025.findings-acl.644/)) and against [`docs-site/docs/algorithms/a2ats.md`](docs-site/docs/algorithms/a2ats.md).

The repo is explicit that this is **"A2ATS-adapted", not a faithful port**, and the doc's "honesty crux" already discloses the proxy-query gap, the calibration footgun, and the missing kernel fusion. Those are **not** what this issue is about. This issue is about four things that are *not* disclosed anywhere, three of which contradict the paper's own equations rather than knowingly adapting them.

Notably, **finding 1 is very likely the mechanical cause of the benchmark result the doc reports as an inherent property of the method** — the "windowed RoPE is worse in every geometry" number in the doc's benchmark section. If that's right, the doc's "When to use it" section is currently steering users away from the method for the wrong reason.

Everything below is reproduced against the checked-in code at `master` (`ed5e2e3`).

---

## Finding 1 — Far tokens are rotated by `R_window`; the paper leaves them **unrotated** (`k̃ = k`)

**Severity: high — this is the paper's central claim, and it is inverted.**

**Where:** [`veloxquant_mlx/quantizers/a2ats_rope.py:133-137`](veloxquant_mlx/quantizers/a2ats_rope.py#L133-L137)

Paper Eq. (11)/(12). WRoPE puts the constant rotation `R_b` on the **query**, and the key for a far token stays in its **pre-RoPE** frame:

> `u_ij = q_i R_b k_j^T`,  for `i − j ≥ w`
> Eq. (12): `q̃_i = q_i R_b`, **`k̃_i = k_i`**

The paper says this in as many words: *"Since post-PE key `k̃_i` states are identical to their pre-PE counterparts `k_i`, WRoPE decouples the positional dependency from post-PE representations."* **That decoupling is the entire reason WRoPE exists** — it is what makes a single shared codebook viable across inputs (Observation 2, §3.1). The whole method is built on far keys being position-free.

The implementation instead rotates the *key* by a fixed `R_window`:

```python
far_offset = mx.array([max(float(window), 0.0)], dtype=mx.float32)
far_cos, far_sin = _rope_cos_sin(far_offset, x.shape[-1], base)   # [1, half]
approx = _rotate(x16, far_cos, far_sin)
```

`q_i R_b k_j^T` and `q_i (k_j R_w)^T = q_i R_w^T k_j^T` are not the same operation — the rotation is on the wrong operand and in the wrong direction. Reproduced:

```python
x = mx.ones((1, 8))
far = a2ats_apply_windowed_rope(x, mx.array([0.0]), query_position=1000, window=64)
# far-token output == input (unrotated)?  False
# far output: [[-0.528, 0.876, 0.205, 0.934, 1.311, 1.109, 1.399, 1.062]]
```

Per Eq. (12) this should return `x` unchanged.

**Two consequences:**

1. **This is very likely why the benchmark says windowed RoPE loses everywhere.** Applying a spurious `R_window` rotation to every far key injects reconstruction error that the paper's formulation never incurs. [`docs-site/docs/algorithms/a2ats.md:170-177`](docs-site/docs/algorithms/a2ats.md#L170-L177) reports ~2.8x MSE on `local_recency` and ~4.4x on `long_range_dependent`, and reads that as an intrinsic cost of the approximation. Under Eq. (12) far keys are untouched, so the `local_recency` penalty in particular should be far smaller than 2.8x. **The benchmark numbers should be re-run after the fix before the doc's conclusion stands.**

2. **The stated motivation for the method is silently defeated.** [`a2ats_rope.py:9-20`](veloxquant_mlx/quantizers/a2ats_rope.py#L9-L20) says WRoPE "decouples the positional dependency," but rotating far keys by `R_window` leaves them in a *rotated* frame. The codebook is then asked to quantize rotated vectors — which is precisely the situation §3.1 identifies as the problem WRoPE was invented to solve.

**Suggested fix** — `approx` becomes the identity, and `R_b` moves to the query side:

```python
approx = x16   # Eq. (12): k̃ = k for far tokens
```

The query-side `R_b` has nowhere to live today (see finding 2).

---

## Finding 2 — The constant `b` is conflated with the window size `w`; the paper's `b` is a separate hyperparameter (`w=64`, `b=2048`)

**Severity: medium — a hyperparameter the paper tunes independently is not expressible.**

**Where:** [`veloxquant_mlx/quantizers/a2ats_rope.py:133`](veloxquant_mlx/quantizers/a2ats_rope.py#L133), config at [`veloxquant_mlx/cache/base.py:238-245`](veloxquant_mlx/cache/base.py#L238-L245)

Eq. (11) has two independent knobs: `w` (the near/far threshold) and `b` (the constant relative position for far tokens). §5.1 Implementation Details sets them to **`w = 64` and `b = 2048`** — different by a factor of 32, so the paper clearly does not treat them as one value.

The code derives the far offset from `window`, hardcoding `b == w`. There is no `a2ats_b` config field; `grep` for `2048` across `cache/base.py`, `a2ats_rope.py`, and `benchmark_scripts/benchmark_a2ats.py` returns nothing. The paper's own reported configuration cannot be expressed.

This also creates a latent trap that a test already had to work around — [`test_windowed_rope_outside_window_uses_fixed_offset`](veloxquant_mlx/tests/quantizers/test_a2ats_rope.py#L91-L110) documents that a far token whose absolute position equals `window` coincides exactly with the far-bucket rotation, and offsets its positions to dodge it. With an independent `b` that coupling disappears.

**Suggested fix:** add `a2ats_b: int = 2048` to `KVCacheConfig`, thread it through `A2ATSKVCache`, and add a `b` parameter to `a2ats_apply_windowed_rope` (defaulting to `2048`, not to `window`). Combined with finding 1, `b` should parameterize the rotation applied to the **query**.

---

## Finding 3 — `query_position` is the last token of the current chunk, and the resulting rotation is frozen into the cache

**Severity: high — near/far classification is wrong for the entire prefill, and never corrected.**

**Where:** [`veloxquant_mlx/cache/a2ats_cache.py:159`](veloxquant_mlx/cache/a2ats_cache.py#L159), `k_out` stored at [`a2ats_cache.py:239`](veloxquant_mlx/cache/a2ats_cache.py#L239)

```python
query_position = int(positions[-1].item()) if S > 0 else self._next_position
```

In Eq. (11), `i` is the position of the **decoding query**, and it advances every step — so a token's near/far class is a function of the *current* decode step. Here `query_position` is `positions[-1]`, the last token of the chunk being written, and the rotated result is written into the parent `KVCache` once and never revisited.

Two distinct problems:

**(a) Prefill misclassifies almost everything.** A 400-token prefill with `window=128` uses `query_position=399`, so positions 272–399 are "near" and 0–271 are "far" — a split computed against a query that does not exist yet. Every subsequent decode step inherits it.

**(b) The rotation never refreshes.** A token classified "near" at prefill keeps its exact rotation forever, even once the decode position has moved thousands of tokens past it. Reproduced with `head_dim=8`, `window=4`, prefill `S=6`, then 10 decode steps:

```
token0 rotation changed as decode advanced?  False
```

Token 0 is >10 steps away and long outside the window, but retains the rotation chosen when `query_position` was 5.

I want to be fair about the constraint here: this is genuinely hard under the mlx_lm cache protocol, since `update_and_fetch` returns rotated keys that the parent cache concatenates and never recomputes. A full fix likely means re-rotating the near-window slice on each decode step, or storing pre-RoPE keys and rotating at fetch. **At minimum this belongs in the doc's honesty crux**, since it silently defeats the distance-gating the method is named for. [`docs-site/docs/algorithms/a2ats.md:140-146`](docs-site/docs/algorithms/a2ats.md#L140-L146) currently describes the intended per-decode-step behavior with no note that it is frozen at write time.

---

## Finding 4 — Query-aware assignment implements cosine-to-centroid, not the paper's `H`-weighted quadratic form

**Severity: medium — documented as an adaptation, but the doc overstates the correspondence.**

**Where:** [`veloxquant_mlx/quantizers/a2ats.py:52-96`](veloxquant_mlx/quantizers/a2ats.py#L52-L96)

Paper §4.2, Eq. (13)/(14): query-aware VQ minimizes a quadratic form weighted by the **second-moment matrix of query states** `H = E[q̃^T q̃]`:

```
f'(k̃; C) = argmin_j (k̃ − c_j) H (k̃ − c_j)^T
```

with `H = LLᵀ` (Cholesky, Eq. 15–19) so that k-means++ runs in the transformed space `z = k̃L`. §3.2's whole argument is that the two objectives align **only if `H ∝ I`**, and Figure 2 shows it is not — that non-diagonal structure *is* the contribution.

The implementation uses a convex blend of normalized reconstruction error and query-to-**centroid** cosine similarity:

```python
score = beta * err_term + (1.0 - beta) * cos_sim[None, :]
return mx.argmax(score, axis=-1).astype(mx.int32)
```

No `H`, no Cholesky — `grep -rn "cholesky\|second_moment"` over `veloxquant_mlx/` returns nothing. This is a different estimator, not an approximation of Eq. (14): `cos_sim` depends only on `(codebook, query)`, so it contributes an **identical per-centroid bias to every row of `x`**, whereas Eq. (14) couples `k̃` and `H` per token. Behavior at `head_dim=2`, centroids `[[1,0],[0,1],[10,10]]`, `x=[[0.9,0]]`, `q=[0,1]`:

```
beta 1.0  -> [0]     # plain nearest-centroid
beta 0.99 -> [0]
beta 0.5  -> [1]     # flips to the query-aligned centroid
beta 0.0  -> [1]
```

Self-consistent, and the `beta=1.0` reduction holds. But `err_term` is normalized per-row by `max_err`, which makes `beta` scale-dependent — the same `beta` means different things for different key magnitudes, which `H`-weighting would not.

The module docstring is careful and honest about the mechanism. The **doc** is not: [`docs-site/docs/algorithms/a2ats.md:133-137`](docs-site/docs/algorithms/a2ats.md#L133-L137) presents the blend as *the* query-aware mechanism without noting the paper's is an `H`-weighted quadratic form. The honesty crux lists five caveats and this omission is arguably the largest.

---

## Non-issues checked and cleared

Ruling these out explicitly so nobody re-investigates:

- **RoPE pairing convention** — `_rotate` half-splits, matching `mx.fast.rope(traditional=False)`; max abs diff `8.7e-4` (fp16 rounding). Correct.
- **`retrieval_fraction=0.0` still returns 1 token** — deliberate `max(1, ...)` guard, covered by [`test_a2ats.py:146-150`](veloxquant_mlx/tests/quantizers/test_a2ats.py#L146-L150). Documented intent, not a bug.
- **`_scatter_1d`** — correct for disjoint indices; verified `[0,7,0,9,0]`. Retrieval/bulk sets are disjoint by construction.
- **Values skipping RoPE** — correct, values are never position-rotated.
- **Proxy query, random-init codebook, no kernel fusion** — already disclosed in the honesty crux.
- **Single proxy query for a whole prefill chunk** — `proxy_query = k_bh[-1]` means one key defines the retrieval set for all 400 prefill tokens. Follows from the disclosed proxy-query limitation; noting it as a sharper edge than the doc implies, not filing it separately.

---

## Suggested resolution

| # | Change | Files |
|---|---|---|
| 1 | Far tokens return **unrotated** keys (Eq. 12); move `R_b` to the query side | [`a2ats_rope.py:133-137`](veloxquant_mlx/quantizers/a2ats_rope.py#L133-L137) |
| 2 | Add `a2ats_b` (default `2048`), independent of `a2ats_window` (`64`) | [`base.py:238-245`](veloxquant_mlx/cache/base.py#L238-L245), [`a2ats_rope.py`](veloxquant_mlx/quantizers/a2ats_rope.py), [`a2ats_cache.py`](veloxquant_mlx/cache/a2ats_cache.py) |
| 3 | Re-rotate the near window per decode step, or store pre-RoPE keys and rotate at fetch. **If infeasible under the mlx_lm protocol, document it in the honesty crux** | [`a2ats_cache.py:159`](veloxquant_mlx/cache/a2ats_cache.py#L159), [`a2ats_cache.py:239`](veloxquant_mlx/cache/a2ats_cache.py#L239) |
| 4 | Either implement `H`-weighted assignment + Cholesky, or state plainly in the doc that the blend is a substitute for Eq. (14) | [`a2ats.py:52-96`](veloxquant_mlx/quantizers/a2ats.py#L52-L96), [`a2ats.md:133-137`](docs-site/docs/algorithms/a2ats.md#L133-L137) |
| 5 | **Re-run `benchmark_a2ats.py` after fix 1** and revisit the "windowed RoPE is worse everywhere" conclusion and the "When to use it" guidance | [`benchmark_scripts/benchmark_a2ats.py`](benchmark_scripts/benchmark_a2ats.py), [`a2ats.md:161-188`](docs-site/docs/algorithms/a2ats.md#L161-L188), [`a2ats.md:240-249`](docs-site/docs/algorithms/a2ats.md#L240-L249) |

### Test coverage gaps

The existing 51 tests pass and are well-written, but they lock in current behavior rather than paper behavior:

- [`test_windowed_rope_outside_window_uses_fixed_offset`](veloxquant_mlx/tests/quantizers/test_a2ats_rope.py#L91) asserts far tokens *differ* from exact RoPE. Under Eq. (12) far tokens equal the **unrotated input** — the assertion needs inverting to that stronger form.
- No test pins `b` independent of `w` (finding 2).
- No test asserts a token's near/far class tracks the advancing decode position (finding 3) — this gap is why the frozen-rotation behavior went unnoticed.

Happy to open a PR for findings 1 and 2, which are self-contained. Finding 3 needs a design call on the protocol constraint, and finding 4 needs a call on whether to implement `H`-weighting or amend the doc.

rajveer43:
Fixes #29.

Implements all four findings from #29, plus the test-coverage gaps and the benchmark re-run that issue asked for.

**One correction to #29 up front:** that issue hypothesized finding 1 was "very likely the mechanical cause" of the benchmark's "windowed RoPE loses everywhere" result, and that the doc's "When to use it" guidance might therefore be steering users away for the wrong reason. **That hypothesis was wrong.** With Eq. (12) implemented correctly the gap not only survives, it widens. Details in the benchmark section below — the near/far breakdown makes the reason unambiguous.

---

## Finding 1 — Far keys are no longer rotated (Eq. 12)

[`a2ats_rope.py`](veloxquant_mlx/quantizers/a2ats_rope.py)

Paper Eq. (12) is explicit that far keys stay pre-RoPE (`k̃_i = k_i`), with the constant `R_b` on the **query** (Eq. 11, `u_ij = q_i R_b k_j^T`). The old code rotated the *key* by a fixed `R_window`, computing `q_i R_w^T k_j^T` — wrong operand, wrong direction.

```python
# before
far_offset = mx.array([max(float(window), 0.0)], dtype=mx.float32)
far_cos, far_sin = _rope_cos_sin(far_offset, x.shape[-1], base)
approx = _rotate(x16, far_cos, far_sin)
return mx.where(near_mask[:, None], exact, approx)

# after — Eq. (12)
return mx.where(near_mask[:, None], exact, x16)
```

Beyond the score being wrong, this restores the property WRoPE exists for: far keys carry no positional information, which is what makes a single shared codebook viable across inputs (§3.1, Observation 2). The old behavior left them in a rotated frame — the exact situation §3.1 identifies as the problem being solved.

New `a2ats_apply_far_query_rope` supplies the query-side half, surfaced as `A2ATSKVCache.far_query_rope`. It is exposed rather than applied internally because `update_and_fetch` never receives a query.

## Finding 2 — `b` is independent of `w`

[`base.py`](veloxquant_mlx/cache/base.py), [`a2ats_rope.py`](veloxquant_mlx/quantizers/a2ats_rope.py)

New `a2ats_b: int = 2048`. The paper's §5.1 uses `w=64`, `b=2048`; deriving the far offset from `window` hardcoded `b == w` and made that configuration inexpressible.

This also removes the latent coupling that [`test_windowed_rope_outside_window_uses_fixed_offset`](veloxquant_mlx/tests/quantizers/test_a2ats_rope.py) previously had to dodge with offset positions — a far token at absolute position `== window` no longer coincides with the far-bucket rotation.

## Finding 3 — Distance gating tracks the decode position

[`a2ats_cache.py`](veloxquant_mlx/cache/a2ats_cache.py)

#29 framed this as fix-or-document because the mlx_lm protocol makes it awkward. A real fix turned out to be available: the parent buffer holds the **pre-RoPE** reconstruction, and windowed RoPE is re-applied to the full accumulated slice on every call against `query_position = self._next_position - 1`.

Tokens now cross from near to far as decode advances, matching Eq. (11) where `i` is the decoding query's position. Previously a token classified "near" at prefill kept its exact rotation permanently.

**Cost, stated plainly:** the rotation is `O(total_tokens)` per step rather than `O(S)`. That is inherent — the near/far split is a function of the current query, so it cannot be precomputed. Far tokens return unrotated, so only the `window`-sized near slice does real work, but the pass itself covers the full cache. Documented in the honesty crux as point 5 rather than buried.

## Finding 4 — The paper's Eq. (14) objective is now implemented

[`a2ats.py`](veloxquant_mlx/quantizers/a2ats.py)

#29 offered "implement `H`-weighting or amend the doc." Implemented, since the paper gives a complete recipe:

- `a2ats_query_second_moment` — `H = E[qᵀq]` (Eq. 10), with a ridge so Cholesky has a positive-*definite* input even when `M < d`
- `a2ats_cholesky_factor` — `L` with `H = LLᵀ` (Eq. 15)
- `a2ats_h_weighted_assignment` — `argmin_j (k̃ − c_j) H (k̃ − c_j)ᵀ` (Eq. 14)

Computed via the Eq. (17) identity, so it is **exact, not approximate**: with `H = LLᵀ`, the `H`-weighted form is a plain Euclidean distance in `z = k̃L`. Verified against a brute-force `einsum` of Eq. (14).

Enabled by the new `a2ats_query_h` config field. Without it the cosine blend still runs — but it is now documented as a *substitute*, not the paper's estimator, with the two concrete differences spelled out (constant per-centroid bias vs. per-token coupling; scale-dependent `beta`).

---

## Benchmark — the #29 hypothesis was wrong

The old benchmark compared reconstructed *keys* against exact-RoPE'd keys. Under Eq. (12) that target is invalid: far keys are *designed* to be unrotated, so measuring them against rotated ground truth scores the method's own design as error. It now measures **attention-score** error `u_ij = q̃_i k̃_j^T`, which is what WRoPE actually approximates and the quantity Eq. (11) is written in.

| Geometry | Bucket | Windowed | Always-exact | Ratio |
|---|---|---:|---:|---:|
| `local_recency` | overall | 3.929 | 1.347 | **2.9x** |
| | near (n=16) | 8.248850 | 8.248850 | **1.000000x** |
| | far | 3.830 | 1.080 | 3.5x |
| `long_range_dependent` | overall | 8.384 | 1.014 | **8.3x** |
| | near (n=16) | 1.090521 | 1.090522 | **1.000001x** |
| | far | 8.762 | 1.006 | 8.7x |

The near/far split settles it. **The near bucket is now numerically identical to always-exact** (max gap `5e-07`) — that is fix 1 working. The entire penalty comes from far tokens, which are ~92% of a 200-token sequence and ~96% of a 400-token one. Replacing each far token's true relative position with one constant `b` is intrinsically lossy; I swept `b` from 8 to 2048 and the gap never closes (smallest at small `b`, growing with it).

So the doc's original "windowed RoPE is worse in every geometry" conclusion stands, and the "When to use it" guidance was right. I sharpened it rather than reversing it: tokens inside the window are now handled *exactly*, so raising `a2ats_window` trades memory savings for fidelity along a predictable curve.

Determinism re-verified by diffing two runs across all non-`ms` fields.

---

## Tests: 51 → 67, full suite 1500 pass

The three gaps #29 flagged, plus the strengthened assertion:

- **`test_windowed_rope_outside_window_returns_key_unrotated`** — replaces the old `_uses_fixed_offset` test. The old one asserted far tokens merely *differed* from exact RoPE, which the buggy `R_window` rotation satisfied, so the bug passed. Now asserts equality with the *input*.
- **`test_windowed_rope_far_tokens_are_position_independent`** — far keys at positions 0/17/133/2040/9001 produce identical output, pinning §3.1's shared-codebook premise.
- **`test_far_query_rope_b_is_independent_of_window`**, **`test_far_query_rope_reconstructs_paper_attention_score`** — `b` is a real separate knob, and the two halves compose back into Eq. (11).
- **`test_token_rotation_updates_as_decode_position_advances`**, **`test_far_tokens_in_cache_are_unrotated`** — near/far tracks the advancing query. This was the gap that let finding 3 go unnoticed.
- **`test_h_weighted_assignment_matches_bruteforce_eq14`**, **`test_h_identity_reduces_to_plain_nearest_centroid`**, **`test_h_weighted_differs_from_plain_vq_under_anisotropic_h`** — the Cholesky route computes Eq. (14) exactly; reduces to plain VQ exactly when `H ∝ I` (§3.2's premise); and genuinely diverges under anisotropic `H`, so the paper's contribution isn't vacuous.

## Docs

[`a2ats.md`](docs-site/docs/algorithms/a2ats.md): corrected mechanism description, honesty crux extended to 9 points (adding the cosine-blend substitution and the `O(total_tokens)` re-rotation cost), new section on calibrating `H` for the Eq. (14) path, benchmark table with the near/far breakdown, updated evidence list, and two new "what we do NOT implement" entries (no automatic `H` calibration; no query-side `R_b` inside the cache).

## Not addressed

- **Proxy query.** `update_and_fetch` still never sees the real query — unchanged, and already disclosed as honesty crux point 1.
- **`H` calibration.** `a2ats_query_second_moment` computes `H` from queries you collect; nothing here collects them. The paper's §4.2 offline pass is the caller's job.
- Everything #29 cleared as a non-issue (RoPE pairing, the `max(1, ...)` retrieval guard, `_scatter_1d`, values skipping RoPE) is untouched.


rajveer43:
Pushed `43c4214` — documentation rewritten from an end-user perspective.

The page previously opened with a mechanism-gap comparison table and a 9-point "honesty crux" before ever saying what the method does or how to run it. That was written for someone auditing the port against the paper, not for someone deciding whether to use it.

**New structure:**

| Section | Purpose |
|---|---|
| Opening | What it does and who it's for, in two sentences, no jargon |
| Quick start | Runnable config with only the fields that matter |
| Should you use this? | Concrete good/poor-fit workloads + pointers to alternatives |
| Tuning `a2ats_window` | The accuracy/savings dial, as a table |
| Calibration | Full walkthrough, matching `vecinfer.md`'s shape |
| Configuration reference | Every field, one table |
| Measuring what you got | The byte-accounting properties, with a usage snippet |
| Honest limitations | 7 numbered items, user-facing terms |
| Benchmark | Same numbers, plus how to read them |
| How it works / Where it sits | Mechanism, moved below the practical material |
| For contributors | Paper fidelity, in a collapsed `<details>` |

**Nothing was softened.** Every caveat survives — the windowing cost, the proxy-query gap, the cosine-blend substitution, the reconstruction-vs-relevance trade (including the explicit "don't read those rows as query-aware being worse" warning), the `O(total_tokens)` per-step cost, and the not-validated-on-a-real-model disclaimer. Benchmark numbers are unchanged; the table now labels buckets "inside/outside window" instead of "near/far" and explains the read.

The main judgment call: paper-equation detail (Eq. 11/12/14, Cholesky, the `k̃_i = k_i` notation) moved into a collapsed "For contributors" block rather than being deleted. It's the right level of detail for reviewing this PR and the wrong level for someone picking a method — a reader who wants it is one click away.

Also picked up house conventions from sibling pages (`:::warning`/`:::note` admonitions, config table, "See also" footer), and refreshed the two `overview.md` entries, which still described the pre-#29 mechanism as a "shared approximate rotation outside it."

**Verified:** every code snippet executed against the real API (calibration, quick start, `a2ats_query_h`, `use_query_aware=False`, and the metrics block); all 11 cross-links resolve; `docusaurus build` passes clean.

rajveer43:
@Som23Git Please review this PR when you get time

rajveer43:
@Som23Git currently the UI is opened locally 
not wired with webUI macos app would be heavy right now so lets ship this first let me know your thoughts

Som23Git (commented):
@rajveer43 Thanks for the pings — reviewed.

### On shipping local web UI first
**Yes — ship the web panel first** (see review on #37). Native macOS can wait. Your note landed on this PR but the decision applies to #37; I'm aligning there.

### On this A2ATS fix (#29)
The code direction looks right:
- Eq. (12) far keys unrotated + `a2ats_b` independent of `w`
- Distance gating re-applied at fetch time
- Eq. (14) H-weighted path with honest cosine-blend docs
- Stronger tests (esp. equality-with-input for far keys — the old test would have passed the bug)
- Benchmark honesty (near/far split, hypothesis corrected)

**Please fix before merge — commitlint is red:**
```
add push   → not conventional (type/subject empty)
fix docs   → not conventional
```
Squash or reword those two (e.g. `docs(a2ats): …` / `chore(a2ats): …`), leave a blank line before any footers, and re-push. Units already pass.

Once commitlint is green I'll approve and merge.

Som23Git (commented):
@rajveer43 Thanks for the pings — reviewed.

### On shipping local web UI first
**Yes — ship the web panel first** (see review on #37). Native macOS can wait. Your note landed on this PR but the decision applies to #37; I'm aligning there.

### On this A2ATS fix (#29)
The code direction looks right:
- Eq. (12) far keys unrotated + `a2ats_b` independent of `w`
- Distance gating re-applied at fetch time
- Eq. (14) H-weighted path with honest cosine-blend docs
- Stronger tests (esp. equality-with-input for far keys — the old test would have passed the bug)
- Benchmark honesty (near/far split, hypothesis corrected)

**Please fix before merge — commitlint is red:**
```
add push   → not conventional (type/subject empty)
fix docs   → not conventional
```
Squash or reword those two (e.g. `docs(a2ats): …` / `chore(a2ats): …`), leave a blank line before any footers, and re-push. Units already pass.

Once commitlint is green **and** the branch is rebased onto current `master`, I'll approve and merge.

Issue #36, opened 2026-07-29
UI: Live KV telemetry & logs — compression, memory, request activity


Som23Git:
## Summary

Parent: [#33](https://github.com/rajveer43/VeloxQuant-MLX/issues/33)  
Companions: [#34](https://github.com/rajveer43/VeloxQuant-MLX/issues/34) Start/Stop · method panel (filed alongside)

This is VeloxQuant's differentiator vs a generic local LLM launcher / stock oMLX: **show whether KV compression is real, measurable, and healthy** while the server runs.

## Why separate

Start/Stop can ship with a health check alone. Live stats need backend hooks from [#27](https://github.com/rajveer43/VeloxQuant-MLX/issues/27) (`/v1/kv/stats` called out there as *the* reason to prefer VeloxQuant over bare `mlx_lm.server`). Bundling charts into MVP would couple the UI to unfinished accounting work.

## Depends on

- [#27](https://github.com/rajveer43/VeloxQuant-MLX/issues/27) exposing honest per-cache / process metrics (`compressed_*_bytes`, `fp16_*_bytes`, peak RSS vs baseline when available)
- [#34](https://github.com/rajveer43/VeloxQuant-MLX/issues/34) running server so the panel has something to poll

### Live stats panel (while Running)
- [ ] Active method + bits
- [ ] Reported compressed vs fp16 byte counters (and ratio)
- [ ] Clear labeling when mode is **accounting-only** (dequantized store) vs **true compressed storage** — never imply RSS savings the backend has not measured
- [ ] Optional: tokens/sec of last request, context tokens in cache

### Logs
- [ ] Scrollable server log (stdout/stderr of the managed process)
- [ ] Filter levels; copy / clear
- [ ] Highlight startup validation failures (unsupported method, OOM, port bind)

### Health / readiness
- [ ] Poll `/health` on an interval; flip status pill in [#34](https://github.com/rajveer43/VeloxQuant-MLX/issues/34) on death
- [ ] If `/metrics` (Prometheus) exists, either link it or show a minimal scrape summary — do not fake it

## Non-goals

- Full Prometheus/Grafana stack
- Historical benchmark suite UI (Throughput / Accuracy sidebar from oMLX can be a later epic)
- Per-layer attention visualizations

## Acceptance criteria

1. With server Running, the panel refreshes without restarting the process.
2. When backend is accounting-only, UI copy says so explicitly (ties to #27 credibility requirement).
[...]

Issue #35, opened 2026-07-29
UI: Method & default-profile panel — understand and pick KV compression without Python


Som23Git:
## Summary

Parent: [#33](https://github.com/rajveer43/VeloxQuant-MLX/issues/33)  
Companion: [#34](https://github.com/rajveer43/VeloxQuant-MLX/issues/34) (Start/Stop lifecycle)

Add the panel that makes VeloxQuant *understandable*: choose among research-adapted KV methods, see plain-language trade-offs, and set method-specific knobs — without opening a notebook.

## Why separate from Start/Stop

oMLX exposes a single built-in KV quantizer flag. VeloxQuant's product is **41 methods** with honest paper deviations. Dumping that into the Server tab crowds the Start/Stop story and hides the education job. This issue owns method discovery + profile editing.

## Problem

Newcomers hit selection paralysis (`turboquant_rvq` vs `kivi` vs `vecinfer` vs eviction methods…). Docs exist on the [docs site](https://veloxquant-mlx.netlify.app/), but they are not in the run loop. The UI should answer:

- What does this method optimize for (memory vs fidelity vs throughput)?
- Recommended default for "local coding assistant" vs "long-context experiment"?
- Which methods are **serve-safe** vs crash / accounting-only per [#27](https://github.com/rajveer43/VeloxQuant-MLX/issues/27) tiers?
- What are the honest deviations from the paper ("-adapted")?

### Method browser
- [ ] List methods available to the running build (drive from registry, not a hardcoded marketing list of "41")
- [ ] Filter / group: quantization vs eviction vs hybrid; serve-tier badge (`serves` / `accounting-only` / `unsupported`)
- [ ] Short blurb + link out to docs-site algorithm page when present
- [ ] Surface "-adapted" caveat when the method documents paper deviation

### Recommended presets
- [ ] At least 2–3 named presets, e.g. `Balanced (RVQ-1bit)`, `KIVI-2bit`, `Max compression (VecInfer-1bit)` — exact set TBD against measured README numbers
- [ ] Preset applies a full `KVCacheConfig`-equivalent (method + bits + relevant budgets)

### Method-specific knobs
- [ ] Show only fields relevant to the selected method (bits / value_bits / budget / seed / artifact path)
- [ ] Validate before Start (refuse unsupported combinations with the same rules as the server)

### Default profile (generation)
[...]

Issue #33, opened 2026-07-29
UX: VeloxQuant control panel (oMLX-like Start/Stop desktop experience)


Som23Git:
## Summary

Ship a first-party **control panel** so people can understand and run VeloxQuant-MLX without writing the 3-line `KVCacheConfig` → `KVCacheBuilder` → `model.make_cache` recipe. The north-star UX is oMLX's Server tab: clear **Start Server / Stop Server**, live status, network settings, copyable API endpoints, and a small set of default generation knobs.

This is **not** \"integrate into oMLX\" ([#5](https://github.com/rajveer43/VeloxQuant-MLX/issues/5)) and **not** LM Studio ([#4](https://github.com/rajveer43/VeloxQuant-MLX/issues/4)). Those put VeloxQuant *inside* someone else's product. This issue is a VeloxQuant-owned surface that *looks and feels* as approachable as oMLX, while teaching what our 41 methods actually do.

## Motivation

Today adoption paths are:

| Path | Friction |
|---|---|
| Python 3-liner + `mlx_lm.generate` | Requires MLX literacy; method choice paralysis across 41 options |
| Docs site / blogs | Great for reading; does not run a local OpenAI-compatible endpoint |
| Future `veloxquant serve` ([#27](https://github.com/rajveer43/VeloxQuant-MLX/issues/27)) | CLI-only unless we add a panel |
| oMLX / LM Studio | Separate products; VeloxQuant is invisible unless [#5](https://github.com/rajveer43/VeloxQuant-MLX/issues/5) / [#4](https://github.com/rajveer43/VeloxQuant-MLX/issues/4) land |

A Start/Stop panel is the missing \"show a friend on Apple Silicon\" demo: install → pick model + method → Start → point Claude Code / OpenAI SDK / Cursor at `http://127.0.0.1:8000/v1`.

## Reference UX (oMLX 0.5.3 Server tab)

Inspired by oMLX's layout (sidebar + main panel):

1. **Header** — app name/version, status pill (`Stopped` / `Running`), primary **Start Server** / **Stop Server** button
2. **Network** — listen address (`127.0.0.1` vs `0.0.0.0`), port (default `8000`), note that save may restart
3. **API endpoints** (copy buttons) — at minimum OpenAI-compatible `…/v1`, health `…/health`; optionally Anthropic-shaped base URL and Prometheus `…/metrics` if the backend exposes them
[...]

Issue #27, opened 2026-07-27
RFC: OpenAI-compatible server — mlx_lm.server already serves 35/40 caches, but compression is accounting-only at runtime


rajveer43:
## Summary

Proposal under discussion: ship an OpenAI-compatible HTTP wrapper inside VeloxQuant-MLX so users can integrate via the OpenAI SDK instead of the 3-line `KVCacheConfig` → `KVCacheBuilder` → `model.make_cache` pattern.

I ran a feasibility pass against v0.41.0 / `mlx_lm` 0.31.3. Two findings invert the premise, and I'd like to settle them **before** any server code is written:

1. **`mlx_lm.server` already serves our caches today.** 35/40 methods subclass `mlx_lm.models.cache.KVCache` and inherit the whole serving contract. A "wrapper" is ~10 lines, not a new subsystem.
2. **But the compression is accounting-only at runtime.** Caches store *dequantized* fp16 tensors, so the inherited `nbytes` over-reports by ~4× vs fp16-raw (and ~16–32× vs true compressed size), and no memory is actually saved.

A server shipped on top of (2) would serve confident, wrong numbers under a banner that implies memory relief. That's a credibility risk to the benchmark work that is this project's core asset — hence this issue rather than a PR.

**Verdict: build-differently.** Fix the memory/accounting story first; the server is a ~10-line follow-on.

Everything below was executed, not inferred. Where I'm inferring or don't know, I say so.

### Finding 1 — the caches already satisfy the serving contract

`mlx_lm.models.cache.make_prompt_cache` defers to `model.make_cache()` when present ([models/cache.py:31-32](https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/models/cache.py#L31-L32)) — exactly the hook `patch_model_kv_cache` overrides. Runtime classification of all 40 methods:

```
SUBCLASS mlx_lm KVCache: 35   (turboquant_rvq, kivi, vecinfer, svdq, palu, xquant, h2o, snapkv, …)
STANDALONE:              5    (turboquant_prod, turboquant_mse, polar, qjl, spectral)
```

Those 35 inherit `update_and_fetch`, `nbytes`, `trim`, `is_trimmable`, `merge`, `meta_state` for free, e.g. `veloxquant_mlx/cache/turboquant_rvq_cache.py:32`:

```python
from mlx_lm.models.cache import KVCache as _MLXKVCache

class TurboQuantRVQKVCache(_MLXKVCache):
```

[...]

PR #37, opened 2026-07-31
feat(ui): VeloxQuant control panel — method registry, OpenAI-compatible serve, Start/Stop web UI

### LINKED ISSUE #34: UI: Server lifecycle — Start/Stop, network, copyable OpenAI endpoints
## Summary

Parent: [#33](https://github.com/rajveer43/VeloxQuant-MLX/issues/33)

Build the **MVP control surface**: one primary action to start and stop a local VeloxQuant-backed inference server, plus the network + endpoint chrome users expect from oMLX's Server tab.

## Why this is its own issue

Start/Stop is the adoption hook. Method education and live KV telemetry are valuable but should not block a working launcher. This issue is the thinnest vertical slice that makes VeloxQuant *runnable* for non-Python users.

## Depends on

- [#27](https://github.com/rajveer43/VeloxQuant-MLX/issues/27) — backend must expose something the UI can spawn (thin `veloxquant serve` → `mlx_lm.server` wrapper at minimum). Until #27 decides (d) vs documented accounting-only launcher, the UI must show an honest status banner matching whatever the server can actually promise.

### Header / lifecycle
- [ ] Status pill: `Stopped` / `Starting` / `Running` / `Error`
- [ ] Primary button: **Start Server** ↔ **Stop Server**
- [ ] Show app/library version (e.g. VeloxQuant-MLX `0.41.0`)
- [ ] Surface subprocess stderr on failure (port in use, unsupported method, model load OOM)

### Network
- [ ] Listen address: `127.0.0.1` (default, safer) and `0.0.0.0` (LAN)
- [ ] Port (default `8000`); changing while running prompts restart
- [ ] Persist settings across launches (local config file)

### API endpoints (copy to clipboard)
- [ ] OpenAI-compatible: `http://127.0.0.1:{port}/v1`
- [ ] Health: `http://127.0.0.1:{port}/health`
- [ ] Include metrics / Anthropic base URL **only if** the backend actually serves them (do not invent endpoints)

### Minimal model bind
- [ ] Model id or local path required before Start (cannot start an empty server)
- [ ] Optional: last-used model remembered

### Default generation knobs (lightweight; deeper method UI is [#35](https://github.com/rajveer43/VeloxQuant-MLX/issues/35) placeholder)
- [ ] Max tokens, temperature, top-p, context window — passed through as server/profile defaults
- [ ] Method + bits fields present but may be a simple dropdown until the method panel lands

## Non-goals

- Full method library browser / paper-deviation callouts → child method UI
- Live compression charts / `/v1/kv/stats` dashboard → child telemetry UI
- Model download manager, appearance themes, security/auth panels
- Reimplementing `mlx_lm.server` inside the UI process

## Acceptance criteria

1. Clicking **Start Server** brings health endpoint to `200` with a small mlx-community model and a #27-supported method.
2. Clicking **Stop Server** terminates the child process; port is free afterward.
3. Copy buttons produce the correct URLs for the configured host/port.
4. Starting with an unsupported method (#27 crash tier, e.g. standalone caches without adapter) fails at Start with a readable error — **no silent fp16 fallback**.
5. README / docs-site getting-started links this as the non-library first-run path once shipped.

## Implementation notes

- Prefer managing a subprocess (`veloxquant serve …` or equivalent) over embedding HTTP in the UI process.
- Keep optional packaging: do not force UI deps on the core PyPI package ([#33](https://github.com/rajveer43/VeloxQuant-MLX/issues/33) stack discussion).

## Reference

oMLX 0.5.3 → Settings → Server: status header, Network (listen + port), API Endpoints list with copy icons, Default Profile knobs.

rajveer43:
## Summary

Ships the first-party control surface from #33: install → pick model + method → **Start Server** → chat via any OpenAI-compatible client, without writing the `KVCacheConfig` → `KVCacheBuilder` → `model.make_cache` recipe.

```bash
veloxquant panel     # local web UI at 127.0.0.1:7860
```

Three layers, each usable on its own:

| Layer | What it adds |
|---|---|
| `veloxquant_mlx/cache/registry.py` | Method catalog with **probed** serve tiers |
| `veloxquant serve` | OpenAI-compatible server with a VeloxQuant cache (#27 option (a)) |
| `veloxquant panel` | Start/Stop web UI (#34 MVP) |

Closes #34. Refs #33, #35, #36, #27.

## Decisions that need review

**1. This ships on #27 option (a), not (d).** #27 recommends "(d), then (a)", and I am proposing the reverse order — with the credibility risk handled in the UI layer rather than by waiting.

The reasoning: blocking the adoption hook on (d) makes it wait for the highest-risk change in the library — one that touches the core whose benchmark numbers *are* the product. The risk (d) protects against is "a server that implies memory relief it cannot deliver," and that is addressable where the claim is actually made. So every surface showing bytes is permanently labelled accounting-only, and no method may claim honest bytes until (d) lands (enforced by a test, not a convention).

This is the load-bearing call in the PR. If you'd rather hold the panel until (d), the launcher and registry still stand alone.

**2. Web UI now, native macOS deferred.** You asked for native, and the plan doc records that decision — but you then asked for web "for now" to see it working. Phases 0 and 1 are identical under both, so this costs nothing: a SwiftUI shell can replace `ui/` without touching the registry or launcher.

**3. `turboquant_prod` — the library default — is in the crash tier.** So `veloxquant serve` defaults to `turboquant_rvq` instead and says so. Flagging because it's surprising, and because it means the documented default has never been serveable.

## Honesty rules

#27 identified the server as exactly where users assume a memory claim without reading caveats. Each rule below maps to a test:

1. **No silent fp16 fallback** — unsupported methods exit non-zero *before* the model loads.
2. **Byte figures are labelled** — accounting-only never renders as memory saved; RSS is never shown, because it isn't measured.
3. **Method lists come from the registry probe**, never a literal "41".
4. **Endpoints render from the handshake only** — no `/health` or `/metrics`, since `mlx_lm.server` doesn't serve them.
5. **`running` requires the handshake**, never a timer — the pill can't claim readiness mid-load.
6. **Requests naming another model are refused**, so the UI never describes a server other than the one that answered.

## Verification

**1527 tests pass** (1506 pre-existing + 21 new) as of the original panel work below; see the update section further down for the current count. Full plan, including what is deliberately *not* automated and why: [`docs/control-panel-test-plan.md`](docs/control-panel-test-plan.md).

Manual end-to-end on `mlx-community/Llama-3.2-1B-Instruct-4bit`:

- Panel spawns server → `starting` → `running` → real completion (`"Hello! How can I assist you today?"`)
- **Stop frees the port** (#34 acceptance criterion 2) — 0 listeners after
- Two methods (`turboquant_rvq`, `kivi`) end-to-end
- Compression confirmed genuinely in the path by instrumenting `make_cache`: `TurboQuantRVQKVCache`, 3536 vs 13312 bytes
- Refusals verified: crash-tier, unknown method, empty model, path traversal, port conflict

## Three defects found only by running it

All in my own integration code, none visible from reading the source. #27's "~10 lines" estimate was optimistic for this reason:

1. **`ModelProvider` reads ~23 fields off `cli_args`.** Fix: harvest mlx_lm's own parser defaults rather than hand-listing them.
2. **MLX arrays are bound to their creating thread.** Loading on main and generating on mlx_lm's worker crashes with `There is no Stream(gpu, N) in current thread`. I was **wrong twice** here — two hypotheses failed before a control run against *stock* `mlx_lm.server` localised it. Fix: hook `_load` so the model lands on the generation thread, as upstream's own `load_default()` does.
3. **`is_batchable` is computed at the end of `_load`.** Overriding the load left it `False`, silently routing every request down the unbatched path — the same crash by another route.

## Known limitations

- **Accounting-only**: no runtime memory reduction until #27 (d).
- One server per panel instance; no auth on the inference server (`0.0.0.0` warns but doesn't protect).
- 2 of 40 methods manually smoke-tested end-to-end; the rest rely on the registry probe plus existing unit tests.
- Live KV telemetry (#36) and the full method browser (#35) are not in this PR.

## Update — plain-language copy pass, a real bug fix, and a docs-site page

The panel worked, but a lot of its text was written for someone reading the
codebase, not someone who just wants to chat with a local model — things like
"accounting-only", raw GitHub issue numbers, `psutil` install errors, and
"telemetry" showed up directly in the UI. This update rewrites that copy
end-to-end, fixes a real bug in the Stop button, and gets the panel's docs
onto the published site where people can actually find them.

**What changed, in plain terms:**

- **Every banner, label, and error message in the panel now speaks to a
  first-time user, not a contributor.** For example: "Accounting-only. Caches
  store dequantized fp16..." became "Heads up: the compression numbers below
  show how well your conversation data compresses — they don't reduce your
  Mac's memory usage yet." Method tags like "accounting-only" / "unsupported"
  became "available" / "not available yet". The Methods table's "Telemetry"
  column became "Compression info", and "not reported" became "no estimate".
  The About page was rewritten from a spec of internal labels into an actual
  welcome-and-explain page.
- **None of the honesty guarantees changed.** Every number that can't be
  measured still says so instead of showing a fake value or a silent zero —
  the wording changed, not the underlying rule. (See "Honesty rules" above;
  all of them still hold and are still tested.)
- **Fixed a real bug: clicking Stop Server could throw an unhandled error.**
  The `/api/stop` route had no error handling, unlike `/api/start` right next
  to it — so a transient hiccup while stopping the server produced a raw
  server error instead of a clean "stopped" state, and the browser showed a
  generic, wrongly-labelled message ("Could not start the server" — even
  though you'd clicked Stop). Both sides are fixed: the backend now returns a
  proper error response instead of crashing, and the button shows the
  correct message for the action you actually took.
- **The control panel's guide is now on the public docs site, not just in
  the repo.** It previously only lived in `docs/control-panel.md`, which
  nobody browsing veloxquant-mlx.netlify.app would ever find. It's now also
  at `docs-site/docs/guides/control-panel.md`, linked from the sidebar and
  from the Quickstart page's "No Python? Start here instead" callout.
- **Small layout fix on the About page** so it reads as a real page (readable
  text width, consistent spacing) instead of stretching edge-to-edge on a
  wide screen.

**Verification:** full test suite passes (1562 tests, up from 1527 — two
existing tests were updated because they asserted the old wording verbatim,
not because behavior changed). Manually re-ran Start → Stop → Start on
`mlx-community/Llama-3.2-1B-Instruct-4bit` with `kivi` to confirm the fixed
Stop button and the new memory card copy.

## Try it

```bash
python3 -m veloxquant_mlx panel
# → model: mlx-community/Llama-3.2-1B-Instruct-4bit, Start Server, copy Base URL

curl http://127.0.0.1:8000/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{"model":"default_model","messages":[{"role":"user","content":"Say hi"}],"max_tokens":20}'
```


rajveer43:
image

rajveer43:
## Update: slice 1 — method browser, per-method knobs, measured memory

Pushed `1f87054`. Turns the panel from a launcher into something that shows what VeloxQuant actually does. **1561 tests pass** (34 new).

### What I added

| Area | Change |
|---|---|
| **Sidebar shell** | `Server · Methods · About`, deep-linkable (`#methods`). The panel was one long column and wouldn't absorb a third section. |
| **Method browser** (#35) | All 40 methods: family chips, search, serve tier, telemetry coverage, `ADAPTED` badges, detail pane with deviation notes + docs link, "Use this method". |
| **Per-method knobs** | KIVI shows `kivi_group_size`; SVDq shows `svdq_rank` et al. Names/types/defaults/optionality derive from `KVCacheConfig` itself. |
| **`--set FIELD=VALUE`** | New on `veloxquant serve`, so knobs work on the CLI too, and panel + CLI share one validation path. |
| **Measured memory** | Server RSS tagged **measured**, beside the accounting-only estimate. |
| **Model picker** | 19 models from the local HF cache; CLIP/BGE/BERT/VLM filtered out. Surfaces what's on disk — still not a download manager. |

### Why — the finding that reshaped the plan

I probed byte-counter coverage across all 35 servable methods before designing this:

```
keys + values : 13
keys only     :  5   ← includes the serve default, turboquant_rvq
not reported  : 17   ← every eviction method
```

**This inverts the naive version of #36.** A compression-ratio panel would be blank or misleading for half the catalog, *silently*. And it's correct behaviour, not a gap: eviction methods drop tokens rather than compress bytes, so a byte ratio isn't meaningful for them. One number cannot describe both families.

So coverage is now probed the same way serve tier is, and three presentation rules join the existing six — each with a test:

7. **Every number states its provenance** — *measured* or *estimate* — in the UI, not just docs.
8. **Absent telemetry says "not reported"** — never `0`, `—`, or a blank that reads as zero.
9. **Key-only ratios are labelled key-only**, never presented as whole-cache.

Rule 8 is the load-bearing one: a zero in a compression panel reads as "no compression", which is a claim nobody measured.

### Two calls I made against my own first draft

**MLX memory counters are withheld, deliberately.** I built them, then saw them read `8 B`. `mx.get_active_memory()` is process-local — it was measuring the *panel*, not the server. Honest but meaningless, and actively misleading beside server RSS under a "measured" tag. The card now states why they're absent; they arrive with #27's `/v1/kv/stats`.

**Memory framing.** #27 forbids *claiming* memory savings — not discussing memory. Showing measured RSS next to the estimate makes the caveat visible rather than asserted, and the note says outright that **RSS will not fall as compression ratios rise**. That's #27's finding rendered as evidence instead of a disclaimer.

### Test plan

**Automated — 34 new, 1561 total.** Full strategy in [`docs/control-panel-test-plan.md`](docs/control-panel-test-plan.md).

| Group | Notable tests |
|---|---|
| Coverage | Locks the 13/5/17 split; asserts the **default method is keys-only**; asserts eviction methods report none; asserts `"not reported"` is never confusable with `0` |
| Field schema | Types/defaults read from the dataclass; `Optional[int]` → nullable; unknown fields tolerated |
| Overrides | Typed parsing, blank-clears-optional, rejects wrong method's knob, rejects unparseable value, reaches the command line |
| Memory | `None` not `0` when absent; MLX withheld *with a reason*; note disclaims comparability |
| Model picker | Never raises; CLIP/BGE/BERT/VLM excluded |
| API | `/api/models`, `/api/memory`, schema + coverage in `/api/methods` |

**Manual, live on `Llama-3.2-1B-Instruct-4bit`:**

1. Started `kivi` with `kivi_group_size=64` **through the panel** — confirmed the override reached the child (`[veloxquant serve] method overrides: kivi_group_size=64`) and inference ran (`"Red."`).
2. Memory card showed **1.0 GB RSS** measured, with both MLX rows stating why they're absent.
3. Methods view: 40 rows, eviction methods showing italic *"not reported"* rather than zeros.
4. Screenshotted Server and Methods views in the running state.

**Not automated, and why:** browser interaction (no JS test infra in this repo), real model loads (multi-GB download would make CI unrunnable), dark theme (headless Chrome starts with empty `localStorage`).

### A bug the screenshots caught

`applyConfig` set the method dropdown without refreshing the description, so a restored config rendered `turboquant_rvq`'s blurb next to the name `kivi`. Fixed in this push. Worth flagging because **no unit test would have caught it** — it took looking at the rendered page, which is the argument for keeping screenshots in the loop.

### Still open

- `/v1/kv/stats` + live telemetry (#36) — slice 2. It now has the presentation rules it needs to be safe to build.
- The three review questions from the PR description stand: **(a)-before-(d)**, web-before-native, and whether the README's documented default should change given `turboquant_prod` has never been serveable.

rajveer43:
image

Som23Git (approved):
@rajveer43 Reviewed — this is ready to ship. Approving and merging.

### On the three review questions

**1. (a) before (d) — agree.** Blocking the adoption hook on real compressed storage (#27 option d) would stall the only surface most non-Python users will ever see. Your approach is the right one: ship the thin launcher, keep every byte figure labelled *accounting-only*, refuse crash-tier methods before model load, and make `HONEST_BYTES` a tier the UI already knows how to render when (d) lands. The presentation rules (measured vs estimate, never show `0` for missing telemetry, key-only labelled as key-only) are exactly what keeps this from becoming a credibility problem.

**2. Web UI now, native macOS later — agree.** Matches what you asked on #30 / what we scoped in #33. A SwiftUI shell can wrap the same `supervisor` + `serve` later without rewriting the registry. Shipping local web first is the right call; a full macOS app would be heavier than we need for the MVP.

**3. `turboquant_prod` vs serve default — fine for now, with one follow-up.** Defaulting `veloxquant serve` to `turboquant_rvq` is correct (crash-tier methods must not be the Start button). Please add a short note in the README / quickstart that the *library* default (`turboquant_prod`) is not yet serveable, so the two defaults don't silently diverge in users' heads. Not a merge blocker.

### What looks especially solid

- Registry probe (not a hardcoded "41") + serve-tier gating
- READY handshake so the pill can't lie mid-load
- Subprocess supervision (panel never does inference itself)
- Control plane hard-bound to loopback
- Slice-1 method browser + per-method knobs driven from `KVCacheConfig`
- Tests locking the honesty rules

Closes the #34 MVP and a big chunk of #35. #36 (`/v1/kv/stats`) can land next on top of the coverage rules you already probed.

Merging.

Som23Git:
Also answering the note you left on #30 about shipping local web first: **yes — ship this.** Native macOS can be a follow-up epic once the panel + serve path is on `master` and we've seen people actually use `veloxquant panel`.

Som23Git (approved):
@rajveer43 Reviewed — ready to ship once rebased onto current `master` (merge is currently blocked by conflicts, likely from the landing/pre-commit merges).

### On the three review questions

**1. (a) before (d) — agree.** Ship the thin launcher with accounting-only labels; refuse crash-tier methods before model load. Don't block the adoption hook on #27 (d).

**2. Web UI now, native macOS later — agree.** Matches your note on #30. SwiftUI can wrap the same supervisor later.

**3. `turboquant_prod` vs serve default — fine for now.** `turboquant_rvq` as serve default is correct. Follow-up (not a blocker): README note that the library default is not yet serveable.

### Solid
Registry probe, READY handshake, loopback control plane, honesty rules in tests, method browser + knobs. Closes #34 MVP; #36 can follow.

Please rebase onto `master` and ping me — I'll merge as soon as CI is green on the rebased tip.

Som23Git:
@rajveer43 Approved on my side, but GitHub reports **merge conflicts with `master`** (likely from the recent landing / pre-commit merges). Could you rebase `feat/control-panel-plan-33` onto current `master` and push? I'll hit merge as soon as CI is green on the rebased tip.

rajveer43:
@Som23Git  Rebased onto current \`master\` (\`9f70068\`, includes the landing revamp and the ruff pre-commit-hook PRs) — force-pushed the linear rebase, no merge commit.

- One real conflict, in \`veloxquant_mlx/__main__.py\`: master's \`else\` branch error message hadn't been updated to list the \`methods\`/\`serve\`/\`panel\` subcommands this branch adds. Resolved in favor of this branch's message (now lists all five commands).
- Ran \`ruff format\` across the tree to pick up the formatting rules introduced by master's new pre-commit hook (#41) — 14 files needed reformatting (whitespace/line-wrap only, no logic changes), committed separately as \`30dfb3f\` so the diff is easy to audit.
- Full test suite green post-rebase: 1562 passed.

Tip should now be conflict-free against \`master\` and pass the \`ruff format --check\` CI job. Will merge once CI reports green on \`30dfb3f\`.

PR #59, opened 2026-08-03
fix: patch_model_kv_cache builds a fresh cache list per call

### LINKED ISSUE #53: P0: patch_model_kv_cache builds one sticky cache list, leaking KV across multiple generate() calls
Sub-issue of #51 (P0 #3: "Fresh caches per make_cache() (match VLM). Invert the sticky-cache test.")

## Problem
`veloxquant_mlx/integration/mlx_lm_patch.py:57-59`:

```python
caches = KVCacheBuilder.for_model(model, config)
model.make_cache = lambda *_a, **_k: caches
```

`caches` is built **once** at patch time and the same list object is returned on *every* subsequent call to `model.make_cache()`. `mlx_lm`'s `generate()` calls `model.make_cache()` fresh at the start of each generation to get a clean cache. With this patch, a second `generate()` call on the same patched model reuses the first call's cache state (whatever KV entries/offsets are left in it), silently leaking context between unrelated generations.

`veloxquant_mlx/integration/mlx_vlm_patch.py` does not have this problem — it rebuilds fresh per call (per the audit).

## Fix
Rebuild a fresh cache list on every call instead of closing over a single instance:

```python
def patch_model_kv_cache(model: Any, config: KVCacheConfig) -> List[Any]:
    def _make_cache(*_a, **_k):
        return KVCacheBuilder.for_model(model, config)
    model.make_cache = _make_cache
    caches = _make_cache()
    print(f"[veloxquant_mlx] Wired {len(caches)} layer cache(s) with {config.method!r}.")
    return caches
```//
(exact shape TBD by implementer — key point is `model.make_cache` must construct new cache objects per invocation, matching `mlx_vlm_patch.py` behavior.)

## Tests
There is reportedly an existing test that pins the *current* sticky/reused-cache behavior — per the audit, that test needs to be inverted to assert fresh caches per call instead of asserting reuse. Find it under `veloxquant_mlx/tests/integration/` (mirror of `test_mlx_vlm_patch.py`) and flip the assertion.

rajveer43:
## Summary
- `patch_model_kv_cache()` (`veloxquant_mlx/integration/mlx_lm_patch.py`) overrode `model.make_cache` with a lambda that closed over a single `caches` list built once at patch time. Every subsequent call to `model.make_cache()` returned that same list object.
- `mlx_lm.generate()` calls `model.make_cache()` fresh at the start of each generation expecting a clean cache. Reusing one list meant a second `generate()` call on the same patched model silently inherited leftover KV state (offsets, cached keys/values) from the first call — a correctness bug that fails silently (wrong/degraded output) rather than crashing.
- `patch_vlm_kv_cache` (`mlx_vlm_patch.py`) already rebuilds fresh per call via a `_make_cache` closure that calls `KVCacheBuilder.for_model()` each time it's invoked. This PR makes `patch_model_kv_cache` match that pattern exactly.

## Changes
- `mlx_lm_patch.py`: `model.make_cache` is now a closure that calls `KVCacheBuilder.for_model(model, config)` fresh on every invocation, instead of hoisting the result out and returning the same object forever.
- `test_mlx_lm_patch.py`:
  - Inverted `test_patch_model_kv_cache_wires_make_cache`, which previously asserted `model.make_cache() is caches` on every call (i.e. it was pinning the sticky-cache bug as expected behavior). It now asserts each call returns a distinct, correctly-sized list.
  - Added `test_patch_model_kv_cache_does_not_leak_across_generate_calls`, which writes to one call's cache and asserts a subsequent call's cache starts at `offset == 0` — a direct regression test for the leak.

Fixes #53

## Test plan
- [x] `pytest veloxquant_mlx/tests/integration/test_mlx_lm_patch.py -v` — all 4 tests pass, including the new leak regression test
- [x] `pytest veloxquant_mlx/tests/integration/` — full integration suite (20 tests) passes, including `mlx_vlm_patch` tests confirming parity with the VLM patch's fresh-cache behavior
- [x] `ruff check` / `ruff format --check` — clean on both changed files

### DIFF
--- a/veloxquant_mlx/integration/mlx_lm_patch.py
+++ b/veloxquant_mlx/integration/mlx_lm_patch.py
@@ -17,12 +17,12 @@ Usage::
 
 from __future__ import annotations
 
-from typing import Any, List
+from typing import Any
 
 from veloxquant_mlx.cache.base import KVCacheBuilder, KVCacheConfig
 
 
-def patch_model_kv_cache(model: Any, config: KVCacheConfig) -> List[Any]:
+def patch_model_kv_cache(model: Any, config: KVCacheConfig) -> list[Any]:
     """Wire a quantized KV cache into an mlx-lm model's generation path.
 
     ``mlx_lm.generate()`` builds its prompt cache via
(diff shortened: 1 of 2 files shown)


rajveer43:
@Som23Git

PR #60, opened 2026-08-03
fix(deps): declare mlx-lm as a required dependency

### LINKED ISSUE #55: P0: mlx-lm used throughout but not declared as a dependency in pyproject.toml
Sub-issue of #51 (P0 #4 / low-hanging-fruit #4: "Declare mlx-lm in pyproject.toml (and demote matplotlib to an extra if easy). Stops install-then-import fail for the real path.")

## Problem
`pyproject.toml` `dependencies` currently lists:
```toml
dependencies = [
    "mlx>=0.18",
    "numpy>=1.26",
    "matplotlib>=3.8",
]
```

`mlx-lm` is not declared anywhere, yet it is imported directly in library code (verified: 35 `import mlx_lm` / `from mlx_lm` occurrences across `veloxquant_mlx/**/*.py`, including `veloxquant_mlx/integration/mlx_lm_patch.py` and inside `KVCacheBuilder.for_model()` in `veloxquant_mlx/cache/base.py:774`). SUPPORT.md and the quickstart docs both assume `mlx_lm` is available.

A user who does `pip install VeloxQuant-MLX` and follows the README/quickstart (which is the primary integration path, per `for_model`/`patch_model_kv_cache`) gets an `ImportError` on first real use, since `mlx-lm` was never pulled in.

## Fix
Add `mlx-lm` to `[project] dependencies` in `pyproject.toml`. Pin a sensible floor version (check current `mlx_lm` API usage, e.g. `mlx_lm.models.cache.make_prompt_cache`, `mlx_lm.load`, `mlx_lm.generate` for the minimum compatible version).

## Optional (bundle if easy, per audit)
`matplotlib` is only used for plotting/benchmark scripts, not core compression logic — consider demoting it to an optional extra (e.g. `[project.optional-dependencies] plots = ["matplotlib>=3.8"]`) so the core install stays lean. Verify with `grep -rn "import matplotlib" veloxquant_mlx/` first to confirm it's not imported from any non-optional code path.

mmaxjr:
Closes #55

## Problem
`mlx-lm` is imported at module level across 35+ files in `veloxquant_mlx` (e.g. `from mlx_lm.models.cache import KVCache` in every cache implementation, plus `KVCacheBuilder.for_model()` in `veloxquant_mlx/cache/base.py`), but it was never listed in `[project.dependencies]` in `pyproject.toml`.

A user who does `pip install VeloxQuant-MLX` and follows the README/quickstart gets an `ImportError` on first real use.

This was already known internally — `.github/workflows/release.yml` had a `pip install mlx-lm` workaround with a comment explaining exactly this gap, and `requirements.local.txt` separately pins `mlx-lm==0.31.3` for local dev.

## Fix
- Added `mlx-lm>=0.31.3` to `[project.dependencies]` (floor version matches the one already validated in `requirements.local.txt`).
- Removed the now-redundant `pip install mlx-lm` workaround (and its explanatory comment) from `release.yml`, since `pip install -e ".[dev]"` now pulls it in.
- Per the issue's optional suggestion: demoted `matplotlib` to a new `plots` extra (`pip install VeloxQuant-MLX[plots]`), since it's only used by standalone scripts in `veloxquant_mlx/benchmarks/` and one lazily-imported method (`DistortionObserver.plot()`, whose docstring already says "Requires matplotlib") — not by any core/non-optional import path. Verified with `grep -rn "import matplotlib" veloxquant_mlx/` that no test or core module imports it eagerly.

## Testing
I wasn't able to run the test suite locally (this environment is Windows; `mlx`/`mlx-lm` only ship wheels for Apple Silicon macOS), so I verified the change by:
- Confirming `pyproject.toml` still parses correctly (`tomllib.load`).
- Grepping all `import matplotlib` / `import mlx_lm` sites across the package to confirm the dependency moves don't affect any core/eagerly-imported code path.
- Confirming no test file imports `matplotlib`.

CI (which runs on the appropriate platform) should validate the rest.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -44,11 +44,6 @@ jobs:
       - name: Install package + dev/release dependencies
         run: |
           pip install -e ".[dev]"
-          # mlx_lm is not in [project.dependencies] but 35 non-test modules
-          # (including veloxquant_mlx/__init__.py's own import chain) import
-          # it at module level, so the package cannot even be imported
-          # without it -- required here for the test suite to collect at all.
-          pip install mlx-lm
 
       - name: Run test suite (release gate)
         run: python -m pytest veloxquant_mlx/tests -q
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,11 +35,14 @@ classifiers = [
 ]
 dependencies = [
     "mlx>=0.18",
+    "mlx-lm>=0.31.3",
     "numpy>=1.26",
-    "matplotlib>=3.8",
 ]
 
 [project.optional-dependencies]
+plots = [
+    "matplotlib>=3.8",
+]
 dev = [
     "pytest>=8.0",
     "pytest-xdist",


.github/workflows/release.yml
```diff
@@ -44,11 +44,6 @@ jobs:
       - name: Install package + dev/release dependencies
         run: |
           pip install -e ".[dev]"
-          # mlx_lm is not in [project.dependencies] but 35 non-test modules
```
rajveer43, line 47: @mmaxjr can you explain me the change here  in `release.yml` file

mmaxjr: Sure! mlx_lm wasn't declared in [project.dependencies] in pyproject.toml, but it's imported at module level by 35 non-test modules (including veloxquant_mlx/__init__.py's own import chain), so the package can't even be imported without it. To work around that gap, release.yml had a manual pip install mlx-lm step (with a comment explaining why) right after pip install -e ".[dev]", just so the release CI wouldn't break.

This PR fixes the root cause instead: it adds mlx-lm>=0.31.3 to dependencies in pyproject.toml, alongside mlx, numpy, and matplotlib. Now pip install -e ".[dev]" installs mlx-lm automatically as a real project dependency, which makes the manual pip install mlx-lm step in release.yml redundant — so I removed those 5 lines.

In short: moved mlx-lm from being a CI-only workaround to being a properly declared package dependency.

rajveer43: Okay

Issue #51, opened 2026-08-03
Audit findings: trust, first-run, and going deeper before more features


Som23Git:
@rajveer43 I did a complete audit of the repo. Sharing thoughts so we can decide what to dig into next.

I think we are shipping features over features without going deep enough on the foundations. Method count and new surfaces keep growing while first-run, honesty of memory claims, and a few core wiring bugs stay open. Happy to be wrong. Here is what I saw.

**What is strong**
Adapted-method honesty in the docs, Metal work, eviction methods that actually shrink tokens, and the validation / packed-storage notes we already wrote. Keep that culture.

**Gaps**
1. **Hero claims vs default path.** A lot of mlx_lm methods still quantize, dequantize, then store fp16, while counters look like big RAM wins. We know this (#27, packed-storage roadmap). Landing / README still read like Activity Monitor will drop. That is a trust risk.
2. **Docs quickstart is broken.** Netlify quickstart still uses `bits=`, `KVCacheBuilder.build(model, ...)`, `kv_cache=`, and MemoryObserver APIs that do not match the real code. README is closer. Newcomers paste the site and fail fast.
3. **Two cache worlds.** mlx_lm `update_and_fetch` vs standalone research caches. "One API for 41 methods" is not true for serving. Default / serve defaults have bitten us already.
4. **Sticky `make_cache`.** `patch_model_kv_cache` reuses one cache list forever. VLM rebuilds fresh. Multi-generate can leak KV. Tests currently pin the sticky behavior.
5. **God config + factory.** Growing `KVCacheConfig` / if-elif factory does not scale. Some "true latent" paths (e.g. KVTC with `_raw_rows`) can still overstate savings.
6. **Packaging / CI.** `mlx-lm` used but not declared; matplotlib hard-dep; PR CI is thin for a Metal library; SUPPORT points at Discussions but Discussions are off.
7. **Product story split.** Panel / serve (#33-#37), LM Studio (#4), oMLX (#5), more methods. We should pick a primary path and close it before adding more catalog surface.

**What we would do next (P0 first)**
1. Fix or take down the broken quickstart / observer docs. Prefer snippet checks in CI.
2. One honest line in hero copy: accounting vs resident RAM. Taxonomy in code: eviction / true_latent / accounting_only / standalone. Refuse incompatible methods in `for_model` / patch.
3. Fresh caches per `make_cache()` (match VLM). Invert the sticky-cache test.
[...]

PR #61, opened 2026-08-03
docs: distinguish accounting vs resident-RAM compression claims

### LINKED ISSUE #54: P0: Hero copy overstates RAM savings — no accounting-only vs resident-RAM distinction
Sub-issue of #51 (P0 #2: "One honest line in hero copy: accounting vs resident RAM. Taxonomy in code: eviction / true_latent / accounting_only / standalone. Refuse incompatible methods in for_model / patch.")

## Problem
README.md hero line (README.md:36):
> "compresses the KV cache of any `mlx_lm` model on Apple Silicon — up to **16× smaller** with near-lossless quality"

This reads as a resident-RAM claim ("Activity Monitor will drop"), but per the audit, most of the 41 methods quantize → dequantize → store fp16 under the hood for the default `mlx_lm` serving path (see #27, packed-storage roadmap). The compression ratio reported by tools like `MemoryObserver` is an *accounting* number (bits actually needed to represent the values), not necessarily the process's peak RSS. Only a subset of methods (eviction-based, or ones with real packed/fused storage) shrink actual resident memory today.

This is a trust risk: landing page and README currently do not distinguish the two, so users benchmark RSS, see no drop, and conclude the library doesn't work.

## Fix (scope for this issue — docs/copy only)
1. Add one explicit line near the hero claim and in the "Numbers that matter" section (README.md:59) clarifying: *"Compression ratios are computed from bit-width accounting. Methods that reduce **resident** memory are marked 🔻RSS in the method table; others reduce precision/quality trade-off at equal storage today (see #27 for the packed-storage roadmap)."*
2. Apply the same line to the Netlify landing page hero (docs-site).
3. In the method library table (README.md:140+), add a column or marker distinguishing method category: `eviction` / `true_latent` / `accounting_only` / `standalone` — matching the taxonomy the audit proposes.

## Out of scope (separate, larger issue)
Adding the taxonomy as an actual enum/field in `KVCacheConfig`/method registry, and using it to **refuse** incompatible methods in `KVCacheBuilder.for_model()` / `patch_model_kv_cache()` with a clear error — track that as its own follow-up since it touches `veloxquant_mlx/cache/base.py` code, not just docs. (Open as a follow-up if not already covered by #27.)

rajveer43:
## Summary
- README hero line and "Numbers that matter" section read as resident-RAM claims ("16× smaller"), but per #27, 35/40 methods subclass `mlx_lm`'s `KVCache` and store *dequantized* fp16 tensors under the hood — the reported compression ratios are bit-width accounting, not measured RSS.
- Adds one explicit clarifying note near the hero claim and above the benchmark table.
- Adds a `Category` column to all three method-library tables classifying each method as `accounting_only` / `eviction` / `true_latent` / `standalone`, per the taxonomy proposed in #51's audit.
- Applies the same disclaimer to the Netlify landing page hero (`landing/index.html`), with a small `.hero-disclaimer` style addition.

## Scope
Docs/copy only, per #54's stated scope. Actually wiring the taxonomy into `KVCacheConfig`/`for_model()` to refuse incompatible methods is explicitly out of scope here and tracked as a separate follow-up.

Fixes #54

## Test plan
- [x] Visual review of rendered README tables (column counts consistent per table, verified via script)
- [x] No Python files touched — ruff/lint/tests unaffected
- [ ] Netlify preview will render on this PR — visually confirm hero disclaimer placement/readability in both light/dark

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,15 @@
 
 **VeloxQuant-MLX** compresses the KV cache of any `mlx_lm` model on Apple Silicon — up to **16× smaller** with near-lossless quality, in three lines of code. It ships **41 research-adapted compression methods**, from zero-calibration 1-bit quantizers to token-eviction caches to cross-layer merging, plus hand-written Metal kernels that make the hottest path **up to 14.7× faster**.
 
+> **Accounting vs. resident memory:** compression ratios above are computed from bit-width
+> accounting (bits actually needed to represent the values), not measured process RSS. Most
+> quantization methods still store dequantized fp16 tensors under the hood on the default
+> `mlx_lm` serving path today, so Activity Monitor / RSS won't drop by the same factor — see
+> [#27](https://github.com/rajveer43/VeloxQuant-MLX/issues/27) for the packed-storage roadmap.
+> Methods that reduce **resident** memory today are marked 🔻RSS in the
+> [method library](#method-library) below (eviction/merging methods, which actually shrink
+> token count); others reduce accounting-only storage at equal fp16 residency.
+
 **Why VeloxQuant-MLX:**
 - 41 methods behind one identical 3-line API — swap `method="..."` and go
 - Metal-accelerated hot paths: 6.9–14.7× faster quantize, 98% less peak memory at the OOM-trigger shape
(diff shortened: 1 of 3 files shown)


rajveer43:
@Som23Git

rajveer43:
@Som23Git I checked. everything is fine

PR #62, opened 2026-08-04
fix: refuse standalone methods in for_model / patch instead of failing deep in generation

### LINKED ISSUE #56: P0: for_model / patch_model_kv_cache silently accept incompatible or crash-tier methods instead of failing loudly
Sub-issue of #51 (P0 #2/#6: "Refuse incompatible methods in for_model / patch with a clear error... Thin check so bad defaults fail loudly instead of later under serve.")

## Problem
`KVCacheBuilder.for_model()` (`veloxquant_mlx/cache/base.py:751`) and `patch_model_kv_cache()` (`veloxquant_mlx/integration/mlx_lm_patch.py:25`) build/wire a cache for whatever `config.method` string is passed, with no validation step that checks whether the requested method is actually safe/complete for the serving path being used (`mlx_lm.generate` autoregressive serving vs standalone research usage).

Per the audit, some methods:
- are only "true latent" savings in specific configurations (e.g. KVTC with `_raw_rows` can overstate savings — see hero-copy issue #54),
- were only validated/tested in the standalone research harness, not under `mlx_lm` serving,
- or have known incomplete/experimental implementations.

Today, picking one of these under `patch_model_kv_cache` for a live `mlx_lm.generate()` call fails silently (wrong output, degraded quality) or fails much later/deeper in generation, rather than at config/patch time — hard to debug for a user just trying the library.

## Fix
1. Add a lightweight per-method compatibility/taxonomy tag (see #54's `eviction` / `true_latent` / `accounting_only` / `standalone` categories — this issue can introduce the enum/registry if #54 hasn't yet) to the method registry used by `for_model`.
2. In `KVCacheBuilder.for_model()` and `patch_model_kv_cache()`, add a validation check up front: if `config.method` is tagged `standalone`-only (not validated for `mlx_lm` serving), raise a clear `ValueError` naming the method and pointing to docs, instead of proceeding.
3. Add a test per entry point asserting the clear error is raised for at least one known-incompatible method.

## Scope note
This depends on/overlaps with #54 (hero-copy taxonomy). If #54's taxonomy lands first, this issue is just "wire the taxonomy into a refusal check at the two entry points."

rajveer43:
## Summary
`KVCacheBuilder.for_model()` and, transitively, `patch_model_kv_cache()` / `patch_vlm_kv_cache()` built and wired a cache for any `config.method` string with no check that the method actually satisfies the `mlx_lm` KVCache serving contract.

Per #27's audit, 5 of the 40 registered methods (`turboquant_prod`, `turboquant_mse`, `polar`, `qjl`, `spectral`) implement VeloxQuant's own `KVCache` ABC (`append_key`/`append_value`/`attend`/`memory_bytes`) instead of subclassing `mlx_lm.models.cache.KVCache` (`update_and_fetch`/`nbytes`/`state`/`trim`/`merge`/`meta_state`). Picking one of these for a live `mlx_lm.generate()` call used to crash or silently misbehave deep inside generation, instead of failing clearly at config/patch time.

## Changes
- `veloxquant_mlx/cache/base.py`: adds `STANDALONE_METHODS` (the 5 methods above) and a refusal check at the top of `KVCacheBuilder.for_model()` that raises `QuantizerConfigError` naming the offending method, before any cache objects are built.
- `patch_model_kv_cache()` / `patch_vlm_kv_cache()` both call `for_model()` before touching `model.make_cache`, so they inherit the refusal automatically — verified with new tests at both entry points.
- Direct standalone/research usage via `KVCacheBuilder.build()` / `KVCacheFactory.create()` is untouched — those remain the documented path for these 5 methods.
- Docstrings updated on both patch functions to document the new `Raises`.
- Cleans up pre-existing `typing.List` → `list` (ruff `UP035`/`UP006`) debt in `mlx_vlm_patch.py`, a file this change edits.

## Scope note
This depends on/overlaps with #54's taxonomy (already landed in #61 as README docs). Per #56's own scope note, since #54 landed first this PR is "wire the taxonomy into a refusal check at the two entry points" using a code-level `STANDALONE_METHODS` set (kept independent of the README table so this PR doesn't depend on #61 merging first).

Fixes #56

## Test plan
- [x] New parametrized test: `for_model()` raises `QuantizerConfigError` for all 5 standalone methods, and still succeeds for a serving-compatible method (`veloxquant_mlx/tests/cache/test_base_for_model.py`)
- [x] New test: `patch_model_kv_cache()` raises before touching `model.make_cache` for `spectral` (`test_mlx_lm_patch.py`)
- [x] New test: `patch_vlm_kv_cache()` raises before touching `language_model.make_cache` for `qjl` (`test_mlx_vlm_patch.py`)
- [x] Full suite: `pytest veloxquant_mlx/tests -q` → 1492 passed (1484 existing + 8 new)
- [x] `ruff check` / `ruff format --check` clean on all touched files

### DIFF
--- a/veloxquant_mlx/cache/base.py
+++ b/veloxquant_mlx/cache/base.py
@@ -8,6 +8,28 @@ from typing import Any, Literal, Optional, Union
 from veloxquant_mlx.core.abstractions import ArtifactStore, KVCache, QuantizationObserver
 from veloxquant_mlx.core.exceptions import QuantizerConfigError
 
+# Methods whose cache class implements VeloxQuant's own KVCache ABC
+# (append_key/append_value/attend/memory_bytes) instead of subclassing
+# mlx_lm.models.cache.KVCache (update_and_fetch/nbytes/state/trim/merge/
+# meta_state). mlx_lm.generate() drives caches purely through the latter
+# interface via model.make_cache() -> cache.update_and_fetch(...), so an
+# object built from one of these methods does not satisfy that contract and
+# either crashes or silently produces wrong output deep inside generation
+# instead of failing at config/patch time. Verified against the codebase in
+# https://github.com/rajveer43/VeloxQuant-MLX/issues/27: these are the 5
+# "standalone" methods among the 40 in the registry; all others subclass
+# mlx_lm's KVCache and are safe to wire into patch_model_kv_cache /
+# KVCacheBuilder.for_model for live mlx_lm serving.
+STANDALONE_METHODS = frozenset(
+    {
+        "turboquant_prod",
+        "turboquant_mse",
+        "polar",
+        "qjl",
+        "spectral",
+    }
+)
+
 
 @dataclass
 class KVCacheConfig:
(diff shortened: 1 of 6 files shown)


rajveer43:
@Som23Git

PR #135, opened 2026-08-10
fix(gear): wire KCVT base quantizer (per-channel keys, per-token values)

### LINKED ISSUE #134: GEAR: base quantizer ignores paper's KCVT axes (per-channel key / per-token value), plus decode-time low-rank overhead exceeds fp16
## Summary

GEAR's base quantizer (`veloxquant_mlx/quantizers/gear.py`) was using a generic per-token group quantizer for **both** keys and values. The paper's actual backbone is **KCVT** (§2, "Group-wise Quantization"): keys quantized **per-channel**, values quantized **per-token** — the same asymmetry [KIVI](../algorithms/kivi) uses, which GEAR builds on directly. This understated the paper's own design and is fixed on the linked branch.

While testing the fix against real models (not just the offline synthetic benchmark), I found a second, previously undocumented issue in how the wrapper amortizes the low-rank correction's overhead.

## Fix 1 — KCVT base axis

`quantize_base` / `gear_compress` / `gear_quant_dequant` now take a `base_axis` (`"channel"` or `"token"`). The cache wrapper (`gear_cache.py`) uses `"channel"` for keys and `"token"` for values. Byte accounting (`gear_bytes`/`base_only_bytes`) is verified axis-independent (same total size regardless of which axis a given tensor uses).

Verified via:
- 9 new unit tests in `test_gear.py` covering axis correctness, the channel-axis-equals-transposed-token-axis identity, and byte-accounting symmetry.
- 1 new cache-level test (`test_cache_uses_kcvt_axes_for_keys_and_values`) asserting the wrapper actually dispatches different axes per tensor kind, not just accepting the parameter.
- Offline synthetic benchmark (`benchmark_scripts/benchmark_gear.py`, regenerated `results_gear.json`) re-run under the KCVT backbone: 11–38% MSE improvement over base-quant-alone across 2/3/4-bit configs, for both axes independently.
- Full repo suite: 1780/1780 tests pass — the change is isolated to GEAR, no collateral impact.

## Finding 2 — decode-time low-rank overhead exceeds fp16 (not fixed, documented)

This is a consequence of an **already-documented** simplification (no streaming buffer — see the module docstring's "Adaptation" section), but its practical impact wasn't previously measured or called out. Surfacing it here because it changes when GEAR is actually worth using.

The low-rank correction costs `(N + D) × rank × 2` bytes, where `N` is however many tokens a given `update_and_fetch` call compresses. At **prefill**, `N` is the whole prompt, so the fixed `D × rank` part of that cost is amortized across hundreds/thousands of tokens. At **decode**, this wrapper compresses **one new token at a time** (`N=1`, no streaming buffer to batch decode tokens the way the paper's `nb=20` buffer does), so the same fixed-size `L`/`R` overhead is paid for a single token — and it is bigger than that token's entire fp16 cost, for any rank ≥ 1. There is no rank that fixes this at `N=1`; it's structural, not a tuning problem.

**Measured on real models** (`mlx_lm.generate`, `gear_bits=2`, `gear_rank=4`, KCVT backbone):

| Model | head_dim | Workload | Key-cache compression vs fp16 |
|---|---|---|---|
| Mistral-7B-Instruct-v0.3-4bit | 128 | ~800 prompt tokens, 8 generated — **prefill-dominated** | **3.96×** |
| Mistral-7B-Instruct-v0.3-4bit | 128 | short prompt, 60 generated — **decode-dominated** | **0.81×** (worse than fp16) |
| Llama-3.2-1B-Instruct-4bit | 64 | short prompt, 60 generated, rank=8 | 0.85× |
| Llama-3.2-1B-Instruct-4bit | 64 | short prompt, 60 generated, rank=2 | 0.92× (best attempted, still <1×) |

**Rule of thumb now documented in `docs-site/docs/algorithms/gear.md`:** GEAR is a net win for prefill-heavy workloads (long-context understanding, summarization, RAG, one-shot classification) and a net loss for decode-heavy generation (long free-form output, extended chat, CoT) — the opposite of what a user would assume from "it's a compression method," so it needed to be stated plainly rather than left implicit in a byte-accounting caveat.

Two regression tests pin this as expected behavior (`test_single_token_decode_call_exceeds_fp16`, `test_prefill_batch_call_can_beat_fp16`) so it can't silently regress further or be "fixed" by accident in a way that hides the real fix (porting the streaming buffer) — see "Not fixed here" below.

## Not fixed here

Porting the paper's streaming buffer (batch `nb=20` decode tokens before applying the low-rank correction, matching KIVI's residual-window pattern already used elsewhere in this repo) would close the decode-time gap. That's a separate, larger change than the KCVT axis fix and is left as follow-up — flagging it explicitly so it's tracked rather than forgotten.

## Files changed

- `veloxquant_mlx/quantizers/gear.py` — `base_axis` plumbing, axis-aware byte accounting
- `veloxquant_mlx/cache/gear_cache.py` — wrapper dispatches `"channel"`/`"token"` per tensor kind
- `veloxquant_mlx/tests/quantizers/test_gear.py` — 11 new tests
- `veloxquant_mlx/tests/cache/test_gear_cache.py` — 1 new test
- `benchmark_scripts/benchmark_gear.py` — benchmarks both KCVT axes separately
- `benchmark_scripts/results_gear.json` — regenerated under the KCVT backbone
- `docs-site/docs/algorithms/gear.md` — rewritten: consumer-friendly explanation, KCVT description, and the prefill-vs-decode warning with measured numbers

rajveer43:
## Summary

- Wires GEAR's base quantizer to the paper's actual **KCVT** backbone: keys quantized **per-channel**, values quantized **per-token** (was: generic per-token quantizer for both).
- Documents a real, measured consequence found while testing against real models: GEAR's low-rank correction is a net win on prefill-heavy workloads and a net loss on decode-heavy generation, because its per-call overhead can't amortize over a single decode token.

Fixes #134

## What changed

- `veloxquant_mlx/quantizers/gear.py` — `base_axis` param threaded through `quantize_base` / `gear_compress` / `gear_quant_dequant`; axis-aware byte accounting (`gear_bytes`/`base_only_bytes`) verified to report the same total size regardless of axis.
- `veloxquant_mlx/cache/gear_cache.py` — wrapper now compresses keys with `base_axis="channel"` and values with `base_axis="token"`.
- `veloxquant_mlx/tests/quantizers/test_gear.py` — 11 new tests: axis correctness, channel-axis-equals-transposed-token-axis identity, byte-accounting symmetry, and the decode-vs-prefill overhead property.
- `veloxquant_mlx/tests/cache/test_gear_cache.py` — 1 new test asserting the wrapper actually dispatches different axes per tensor kind.
- `benchmark_scripts/benchmark_gear.py` + `results_gear.json` — benchmarks both KCVT axes separately, re-run under the new backbone.
- `docs-site/docs/algorithms/gear.md` — rewritten for a broader audience: plain-language explanation of the three-part decomposition, config table, and a prominent warning with measured prefill-vs-decode numbers so nobody discovers the regression in production.

## Testing

- Full repo suite: **1780/1780 tests pass** — the change is isolated to GEAR, no collateral impact on other methods.
- Offline synthetic benchmark re-run under the KCVT backbone: 11–38% MSE improvement over base-quant-alone across 2/3/4-bit configs (see `results_gear.json`), for both key (channel) and value (token) axes independently.
- Real end-to-end generation via `mlx_lm.generate` on two models — numbers posted as a comment on this PR.

## Not in scope for this PR

Porting the paper's streaming buffer (batching `nb=20` decode tokens before applying the low-rank correction) would close the decode-time overhead gap this PR documents but doesn't fix. Left as follow-up — see #134.


rajveer43:
## Real-model benchmark numbers

Ran `mlx_lm.generate` end-to-end with the KCVT-fixed GEAR (`gear_bits=2`, `gear_group_size=32`) against real models to check the fix didn't just look good on synthetic data, and to quantify the prefill-vs-decode overhead behavior called out in #134.

### Mistral-7B-Instruct-v0.3-4bit (head_dim=128, matches the paper's own setup)

| Workload | Prompt tokens | Generated tokens | `gear_rank` | Key-cache compression vs fp16 |
|---|---|---|---|---|
| Prefill-dominated | ~800 | 8 | 4 | **3.96×** |
| Decode-dominated | short | 60 | 8 | 0.86×* |
| Decode-dominated | short | 60 | 4 | 0.81× |

\* from an earlier run at `gear_bits=2, gear_rank=8` — included for the rank comparison.

### Llama-3.2-1B-Instruct-4bit (head_dim=64 — smaller head dim, less room to amortize)

| Workload | Prompt tokens | Generated tokens | `gear_rank` | Key-cache compression vs fp16 |
|---|---|---|---|---|
| Decode-dominated | short | 60 | 8 | 0.85× |
| Decode-dominated | short | 60 | 4 | 0.90×** |
| Decode-dominated | short | 60 | 2 | 0.92× |

\*\* `gear_bits=4` config in this run, others `gear_bits=2` — noted since bit-width also shifts the base_only floor.

### Takeaways

- The core fix (KCVT axes) doesn't change *whether* GEAR helps — it changes what the base quantizer actually reconstructs (per-channel key structure vs. a generic per-token quantizer). Confirmed correct via the offline synthetic benchmark (`results_gear.json`): 11–38% MSE improvement over base-quant-alone across 2/3/4-bit configs, for both axes.
- The prefill/decode split is the more consequential finding for users: **3.96× on a prefill-heavy call, <1× (worse than fp16) on every decode-heavy configuration tried, regardless of rank.** Lowering rank narrows the gap slightly (0.85× → 0.92× as rank drops from 8 → 2) but never crosses 1× at `N=1`, which matches the structural argument in #134 (`(N + D) × rank × 2` bytes of overhead can't be amortized when `N=1`).
- Output text stayed coherent in every configuration tested — this is a memory/compression-ratio finding, not a correctness or quality regression.

All numbers above are from live model runs (not synthetic), generated while validating this PR.

PR #138, opened 2026-08-10
fix(h2o): RoPE desync, prefill crash, fused Metal kernel, freeze fix, and score-decay fix

### LINKED ISSUE #137: H2O-adapted cache: eviction corrupts real-model generation (RoPE bugs fixed; dominant cause is early-token scoring freeze, not fixed)
## Summary

Real end-to-end generation testing (`mlx_lm.generate` on `Llama-3.2-1B-Instruct-4bit` and `Mistral-7B-Instruct-v0.3-4bit`) found that `method="h2o"` (H2OKVCache, arXiv:2306.14048-adapted) corrupts generation quality as soon as eviction actually fires — as few as one or two evictions can send output into repetition loops, and heavier eviction causes total collapse (single-token loops, blank output).

The investigation found and fixed two real, independently-verified correctness bugs, but also found the **actual dominant cause** is a structural limitation of the paper's own scoring formula, not fixed in this issue.

## Fixed in this PR

1. **RoPE position desync on eviction.** Stored keys arrive pre-rotated (by `mlx_lm`'s attention module) for their true absolute position. Evicting a row from the interior of the kept set left survivors' storage index out of sync with their baked-in rotation, scrambling relative-position math for every subsequent query. Fixed by de-rotating and re-rotating kept keys to a gap-free contiguous layout on every eviction (`rope_remap_positions`, a new primitive in `veloxquant_mlx/quantizers/a2ats_rope.py`).

2. **`cache.offset` conflated with kept-row-count.** `mlx_lm`'s attention module rotates the query *and* the next incoming key using `self.rope(x, offset=cache.offset)` — entirely inside the model's forward pass, before `update_and_fetch` is ever called. `H2OKVCache` previously reset `offset` to `n_kept` every call (the same pattern used by the other 16 eviction/compression caches in this repo — see "Also found" below). Once eviction makes `n_kept < true elapsed steps`, this silently rotates all future queries/keys at the wrong absolute position. Fixed by making `H2OKVCache.offset` track the true absolute step count directly, managing `self.keys`/`self.values`/`self.offset`/`.state`/`.size()` itself instead of delegating to the base class's append-only buffer.

Both fixes are verified via a synthetic fingerprinted-token test that forces an interior eviction and confirms every surviving key de-rotates back to its exact original pre-rotation value (`test_interior_eviction_gap_remap_recovers_exact_original_keys`).

## Root cause of the collapse — NOT fixed in this PR

Despite the two fixes above being real and correct, they turned out **not** to be the dominant cause of the collapse observed on real models. Precise tracing (`H2OState.positions`, `next_pos`) showed that interior eviction essentially never happens in practice:

- H2O's scoring rule gives every newly-arrived token a starting score of exactly `0.0`.
- Eviction removes whichever token holds the **global minimum** score.
- On any realistic (non-degenerate) score distribution, the brand-new token's `0.0` *is* that global minimum — so **the most recently generated token is evicted on almost every step once the cache is full**, before it ever accumulates attention mass.

Traced directly on the Llama-3.2-1B run: after **43 true decode steps**, the cache's kept positions were still exactly `[0..13]` — the original prompt, unchanged, with **zero** generated tokens ever admitted. The model generated the entire time attending only to the prompt, with no visibility into its own output — exactly the repetition-loop failure pattern observed.

This is a genuine limitation of the paper's own cumulative-sum scoring formula at tight/realistic budgets — implementing Algorithm 1 correctly (verified line-by-line against the paper) is what produces this, not a deviation from it. See the "Fidelity to Algorithm 1" table in `docs-site/docs/algorithms/h2o.md`.

**A real fix needs a different eviction/scoring policy** — e.g. a grace period before a token becomes eviction-eligible, or age-normalized scoring — a bigger, more opinionated algorithmic change than a bugfix, and intentionally not attempted here.

## Also found: likely shared bug across 16 other eviction caches

The existing regression test `test_issue_83_state_writethrough.py` asserts `cache.offset == shape[2]` (stored row count) for **all 17** eviction/compression cache methods in this repo. All of them follow the same reset-then-`super().update_and_fetch()` pattern H2O used before this fix. That means StreamingLLM-adapted, SnapKV-adapted, CaM-adapted, KVzip-adapted, MorphKV-adapted, and the other budget-capped caches likely share the exact same query/key-rotation bug fixed here for H2O — **not yet investigated or fixed**. StreamingLLM/SnapKV may be less exposed in practice since their eviction patterns are contiguous-from-one-end (sink + trailing window) rather than scattered, which can keep relative deltas intact even without explicit remapping — but this has not been verified for any of them.

Tracked as follow-up scope, deliberately not investigated in this PR to keep it focused.

## Practical takeaway

**Do not use `method="h2o"` for real generation today.** The eviction mechanism (scoring, sink protection, budget enforcement) is validated as a correct implementation of the paper's Algorithm 1, and the two RoPE correctness bugs above are now fixed — but the cache still freezes on the initial prompt once the budget fills, making it unsuitable for actual model output until the scoring/eviction policy itself is revised.

## Evidence

- 7 new/updated unit tests in `veloxquant_mlx/tests/quantizers/test_h2o.py` and `veloxquant_mlx/tests/cache/test_h2o_cache.py`, plus updated exceptions in `test_issue_83_state_writethrough.py` for H2O's new (correct) offset contract.
- Full suite: 1786/1786 passing.
- Real end-to-end generation on Llama-3.2-1B (`rope_theta=500000`) and Mistral-7B-v0.3 (`rope_theta=1000000`) — exact before/after outputs documented in `docs-site/docs/algorithms/h2o.md`.
- `docs-site/docs/algorithms/h2o.md` rewritten with a `:::danger` section, the "Fidelity to Algorithm 1" comparison table, and the precise root-cause writeup.

rajveer43:
## Summary

Fixes #137 — real end-to-end generation testing found `method="h2o"` corrupts output once eviction fires. This PR fixes two real, independently-verified RoPE correctness bugs, and documents (but does not fix) the actual dominant cause, which needs a bigger algorithmic change.

## What's fixed

1. **RoPE position desync on eviction** — stored keys arrive pre-rotated for their true absolute position; dropping an interior row left survivors' storage index out of sync with that rotation. Fixed via a new de-rotate/re-rotate primitive, `rope_remap_positions` (`veloxquant_mlx/quantizers/a2ats_rope.py`), applied in `h2o_update` whenever eviction changes the kept set.
2. **`cache.offset` conflated with kept-row-count** — `mlx_lm` rotates the query *and* the next incoming key using `cache.offset`, inside the model's forward pass, before `update_and_fetch` runs. `H2OKVCache` previously reset `offset` to `n_kept` every call, silently rotating future queries at the wrong position once eviction occurred. Fixed by tracking `offset` as the true absolute step count, managing `keys`/`values`/`offset`/`.state`/`.size()` directly instead of delegating to the base class.

Verified via a synthetic fingerprinted-token test that forces an interior eviction and confirms every surviving key de-rotates back to its exact original pre-rotation value.

## What's NOT fixed (see #137 for full writeup)

Precise tracing showed interior eviction essentially never happens in practice: every new token starts at score `0.0`, and eviction removes the global minimum, so the newest token is (almost) always its own eviction target. Traced on Llama-3.2-1B: after 43 real decode steps, the kept set was still exactly the original 14-token prompt — the model generated the whole time with zero visibility into its own output, which is the actual cause of the repetition-loop collapse seen on real models. This is a consequence of the paper's own scoring formula, not a bug in this implementation (verified against Algorithm 1 line-by-line — see the new "Fidelity to Algorithm 1" section in the docs). Fixing it needs a different eviction/scoring policy, tracked as follow-up in #137.

Also flagged in #137, not investigated here: the other 16 eviction/compression caches in this repo share the same pre-fix `offset == n_kept` pattern and may have the same latent query-rotation bug.

## Testing

- `veloxquant_mlx/tests/quantizers/test_h2o.py` — 6 new tests (position tracking, contiguity under stress, interior-eviction remap correctness, early-token-freeze documentation).
- `veloxquant_mlx/tests/cache/test_h2o_cache.py` — 3 new tests (`offset` tracks true position, `.state` returns exactly kept rows, `.size()` != `.offset`).
- `veloxquant_mlx/tests/cache/test_issue_83_state_writethrough.py` — updated to carve out H2O's new (correct) `offset != shape[2]` contract as a documented exception, with H2O-specific counterpart assertions.
- Full suite: **1786/1786 passing**.
- `docs-site` build: clean, no broken anchors.
- Real end-to-end generation re-verified on `mlx-community/Llama-3.2-1B-Instruct-4bit` (`rope_theta=500000`) and `mlx-community/Mistral-7B-Instruct-v0.3-4bit` (`rope_theta=1000000`) after the fix — exact outputs documented in `docs-site/docs/algorithms/h2o.md`.

## Docs

`docs-site/docs/algorithms/h2o.md` rewritten: top-of-page `:::danger` banner, a "Fidelity to Algorithm 1" comparison table against the paper, and a precise root-cause section replacing the earlier (overstated) RoPE-only explanation with the verified early-token-freeze finding.

### DIFF
--- a/docs-site/docs/algorithms/h2o.md
+++ b/docs-site/docs/algorithms/h2o.md
@@ -20,7 +20,10 @@ cumulative-sum formulation at low budgets, not an implementation bug in the
 eviction *rule* itself (which correctly implements Algorithm 1 — see
 [fidelity check](#fidelity-to-algorithm-1)). See
 [the finding](#real-end-to-end-generation-eviction-breaks-coherence) before
-using `method="h2o"` for anything beyond studying the mechanism.
+using `method="h2o"` for anything beyond studying the mechanism. A separate
+scalability bug that used to crash long prefills outright (regardless of
+eviction settings) has since been fixed — see
+[long-context testing](#long-context-testing-the-freeze-holds-at-scale-plus-a-scalability-bug-now-fixed).
 :::
 
 H2O-adapted is the library's **third eviction axis** and the first based on
@@ -117,8 +120,8 @@ specified is what causes the practical problem, not a deviation from it.
 ## Evidence
 
 All claims trace to passing tests in
-`veloxquant_mlx/tests/cache/test_h2o_cache.py` (15 tests) and
-`veloxquant_mlx/tests/quantizers/test_h2o.py` (18 tests):
+`veloxquant_mlx/tests/cache/test_h2o_cache.py` (18 tests) and
+`veloxquant_mlx/tests/quantizers/test_h2o.py` (28 tests):
 
 - `init_h2o_state` fields correct; empty state returns zero-row K/V placeholder
 - Single token bootstraps state; multi-token absorption below budget keeps all tokens
@@ -135,6 +138,11 @@ All claims trace to passing tests in
 - Factory dispatch (`KVCacheFactory.create`) returns `H2OKVCache`
 - `for_model` propagates `h2o_budget` and `h2o_n_sink` to all layer caches
 - Determinism: identical inputs produce identical outputs
+- Position tracking stays gap-free and contiguous under a 60-step stress test; `n_sink` positions never move
+- Interior-eviction RoPE remap recovers each surviving key's exact original pre-rotation value (fingerprinted-token test)
+- `cache.offset` tracks the true absolute step count, not the kept-row count, once eviction has occurred
+- The vectorized below-budget batch path matches the sequential per-token loop it replaces bit-for-bit within fp16 rounding, including for a single call whose batch straddles the below-budget/over-budget boundary
+- A 4,000-token single-call absorption (no eviction) completes without error — the regression test for the prefill scalability crash
 
 The offline harness in `benchmark_scripts/benchmark_h2o.py` sweeps
 `(seq_len, budget, n_sink)` and reports latency and compression ratio —
@@ -213,6 +221,62 @@ change to the eviction/scoring policy itself (e.g. a grace period before a
 token becomes eviction-eligible, or age-normalized scoring) — a bigger,
 more opinionated algorithmic change than a bugfix, and not undertaken here.
 Tracked as follow-up work, not silently patched over.
+
+### Long-context testing: the freeze holds at scale, plus a scalability bug (now fixed)
+
+The same methodology was repeated on a genuinely long prompt (~3,238 tokens)
+to check whether the early-token freeze above behaves differently at scale,
+and to stress-test the implementation on long context in general.
+
+| Budget | Prefill | Decode | Kept (final) | Evicted | Compression | Generated tokens survived? | Output |
+|---|---|---|---|---|---|---|---|
+| 256 | 3,238 | 300 | 256 | 3,282 | 13.82× | **0** | `"of of of of of..."` — total collapse |
+| 800 | 3,238 | 300 | 800 | 2,738 | 4.42× | **0** | Garbled/gibberish tokens |
+| 1,200 | 3,238 | 400 | 1,200 | 2,438 | 3.03× | **0** | Degenerates into `"I am I am I am..."` loop |
+
+In every case the kept set is exactly `[0, budget-1]` — the earliest slice of
+the **prompt** — confirming the freeze is budget- and prompt-length-independent:
+bigger budgets don't fix it, they just delay how quickly the output degrades
+from coherent-looking phrases into pure repetition.
+
+:::warning[A second, independent bug was found and fixed here: H2O couldn't even complete a long prefill]
+Testing the no-eviction baseline (`h2o_budget` larger than the prompt, so no
+eviction should ever occur) crashed **during prefill alone**, before a single
+decode step:
+
+```
+RuntimeError: [metal::malloc] Resource limit (499000) exceeded.
+```
+
+Bisected the breaking point: prefill succeeded at 500 tokens, failed
+somewhere before 1,000 — independent of `h2o_budget`, `h2o_n_sink`, or
+`max_tokens` (reproduced even at `max_tokens=1`).
+
+**Root cause:** `h2o_update` processed every incoming token, including the
+entire prefill batch, with a Python `for` loop issuing 4 separate
+`mx.concatenate` calls per token (keys, values, scores, positions). At
+~3,238 tokens this builds an unfused lazy-evaluation graph large enough to
+exceed MLX's Metal resource/command-buffer tracking limit — a pure
+implementation scalability bug, completely independent of the RoPE fixes and
+the scoring-freeze issue above.
+
+**Fixed:** whichever leading portion of an incoming batch is guaranteed not
+to trigger eviction (because the cache hasn't yet reached `h2o_budget`) is
+now absorbed via a single batched masked-attention matmul instead of a
+per-token loop — mathematically the same score-accumulation formula, just
+computed all at once. Verified numerically equivalent to the sequential loop
+it replaces (exact match to fp16 rounding, including for calls that straddle
+the below-budget/over-budget boundary). The same 3,238-token no-eviction
+prefill that previously crashed now completes in ~2.4–3.2 seconds.
+
+Only the genuinely sequential part — the eviction decision itself, where
+each eviction depends on the previous step's result — still runs a per-token
+loop, and only once the cache is actually full. This means the fix has no
+effect on the freeze-heavy scenarios in the table above (their timing and
+output are byte-for-byte identical before and after this fix); it only
+unblocks the case that used to crash outright: long context with a budget
+large enough to avoid heavy early eviction.
+:::
 
 ## When to use it
 
--- a/docs-site/docs/algorithms/h2o.md
+++ b/docs-site/docs/algorithms/h2o.md
@@ -4,6 +4,25 @@
 (Zhang et al., ICLR 2024) — **H2O-adapted (VeloxQuant-MLX implementation)**,
 not a faithful port.
 
+:::danger[Not safe for real generation yet — cache freezes on early tokens]
+Real-model testing (below) confirms eviction corrupts output — as few as one
+or two evictions can send generation into repetition loops, and heavier
+eviction causes total collapse. Root cause: the scoring formula gives every
+newly-arrived token a starting score of exactly `0`, and the eviction rule
+removes the *global minimum* score — so once the cache is full, the brand-new
+token is (almost always) its own eviction target, before it ever gets a
+chance to accumulate attention mass. In practice this means **the kept set
+freezes on whichever tokens filled the budget first** (typically the prompt)
+and never admits anything generated afterward — the model ends up generating
+forever while attending only to the original prompt, with zero visibility
+into its own recent output. This is a real limitation of the paper's own
+cumulative-sum formulation at low budgets, not an implementation bug in the
+eviction *rule* itself (which correctly implements Algorithm 1 — see
+[fidelity check](#fidelity-to-algorithm-1)). See
+[the finding](#real-end-to-end-generation-eviction-breaks-coherence) before
+using `method="h2o"` for anything beyond studying the mechanism.
+:::
+
 H2O-adapted is the library's **third eviction axis** and the first based on
 **cumulative per-token attention mass**. Unlike SnapKV-adapted (which fires once
 at prefill end) and StreamingLLM-adapted (which evicts by position), H2O runs
@@ -61,12 +80,26 @@ For every incoming token (both prefill and decode), per head:
 No `.bits` attribute — stored K/V remain in fp16. The `compression_ratio` and
 `tokens_kept` properties report the storage accounting.
 
-## Proxy limitation
+## Fidelity to Algorithm 1
 
-The paper accumulates attention weights from the **true query** vectors at each
-decode step. At the cache-wrapper level, queries are not visible — only K and V
-arrive at `update_and_fetch`. We substitute the incoming **key vector** as a proxy
-query, computing an approximation of the attention distribution over stored keys.
+Checked directly against the paper's Algorithm 1 / Definition 4.3 (local
+greedy heavy-hitter eviction):
+
+| Paper | This implementation | Match |
+|---|---|---|
+| `Si ← (Si−1 ∪ {i}) \ {u}`, `u ← argmax_v Fscore(Si−1∪{i}\{v})` | Evict the single lowest-scoring element (`argmin`) — equivalent to maximizing the *remaining* set's score | ✅ |
+| `Fscore(T) = Σ_{s∈T} o_s`, accumulated per token | Running sum of softmax weights per token | ✅ |
+| §4.1: local (no-lookahead) H2 is "equally effective" as the global variant — the paper's own recommended, deployable version | Implements exactly the local variant | ✅ (correct choice) |
+| `\|Si\| = k`, evict at most one per step | `budget` field, one eviction per over-budget call | ✅ |
+| §5.3 sink/StreamingLLM extension | `n_sink` protected leading positions | ✅ |
+
+**Proxy limitation:** the paper accumulates attention weights from the
+**true query** vectors at each decode step. At the cache-wrapper level,
+queries are not visible — only K and V arrive at `update_and_fetch`, and (as
+documented in [How it works](#how-it-works) above) `mlx_lm`'s attention
+module consumes the true query internally before the cache is ever called.
+We substitute the incoming **key vector** as a proxy query, computing an
+approximation of the attention distribution over stored keys.
 
 This is the same key-as-query approximation used by SnapKV-adapted. Keys and queries
 are both projected from the same residual stream and are correlated, but the proxy is
@@ -75,6 +108,11 @@ similar to recent keys rather than those that answer the actual query.
 
 Documented as "H2O-adapted (key-as-query proxy)" throughout — never claimed as a
 faithful port.
+
+**Not a fidelity gap, but a consequence of the formula itself:** the paper's
+own scoring rule (new tokens start at score 0, minimum gets evicted) is what
+produces the early-token freeze described below — implementing it exactly as
+specified is what causes the practical problem, not a deviation from it.
 
 ## Evidence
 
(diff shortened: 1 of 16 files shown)


rajveer43:
## Scope extended: also fixes a prefill scalability crash found during long-context testing

Follow-up long-context testing (posted on #137) found a **third**, independent bug beyond the two RoPE fixes this PR started with: `h2o_update`'s per-token Python loop (4 separate `mx.concatenate` calls per token, even during prefill) exceeded MLX's Metal resource limit on long prompts, crashing before generation could even start — **regardless of eviction/budget settings**.

### The crash

```
RuntimeError: [metal::malloc] Resource limit (499000) exceeded.
```

Reproduced with a ~3,238-token prompt and `h2o_budget` set high enough that no eviction should ever occur. Bisected: prefill succeeds at 500 tokens, fails somewhere before 1,000 — independent of `max_tokens` (fails even at `max_tokens=1`).

### The fix

Added `_batch_absorb_no_eviction` / `_batch_absorb_prefix` in `veloxquant_mlx/quantizers/h2o.py`: whichever leading portion of an incoming batch is guaranteed not to trigger eviction (because the cache hasn't yet reached `h2o_budget`) is now absorbed via a single batched masked-attention matmul instead of a per-token loop. The over-budget eviction loop is untouched — each eviction depends on the previous one, so it can't be batched without changing which token gets evicted.

**Verified numerically equivalent** to the sequential loop it replaces:
- Exact match (fp16 rounding only) for pure below-budget absorption, both from empty state and with prior tokens already present.
- Exact match for a single call whose batch straddles the below-budget/over-budget boundary (mixed batch-path + loop-path in one call).

### Before/after on the same long-context scenarios

| Scenario | Before | After |
|---|---|---|
| budget=5000 (no eviction), 3238-token prompt | **Crash** (`Resource limit exceeded`) | **4.29s**, coherent on-topic output |
| budget=256 (heavy eviction) | 240.5s, `"of of of of..."` | 240.5s, `"of of of of..."` — byte-identical |
| budget=800 (heavy eviction) | 208.6s, garbled tokens | 208.6s, garbled tokens — byte-identical |
| budget=1200 (heavy eviction) | 199.7s, `"I am I am..."` loop | 199.7s, `"I am I am..."` loop — byte-identical |

The eviction-heavy scenarios are unaffected in both timing and output — confirming this is a pure performance/scalability fix with zero behavior change to the (already correct) eviction logic, and confirming the early-token-freeze issue tracked in #137 is untouched and still open.

### Testing

- 4 new tests in `test_h2o.py`: batch-vs-sequential equivalence (from empty, with prior tokens), single-call batch/loop-boundary equivalence, and a 4,000-token no-eviction regression test for the crash itself.
- Full suite: **1790/1790 passing** (up from 1786).
- `docs-site/docs/algorithms/h2o.md` updated with the long-context table and a new "Long-context testing" section.

rajveer43:
## Scope extended again: fused Metal kernel for the over-budget eviction step

Added a hand-written Metal kernel pair that fuses `h2o_update`'s per-token eviction branch — previously ~15+ small MLX ops per token (4x `mx.concatenate`, `mx.argmin`, 4x boolean-index, RoPE remap) — into two GPU dispatches.

### What was added

- `veloxquant_mlx/metal/src/h2o_evict_reduce.metal` — sink-protected argmin over the candidate rows, one threadgroup per `(batch*head)` group, using a SIMD-group butterfly reduction (`simd_shuffle_xor`) + threadgroup-memory merge (same idiom as the existing `scalar_affine_attend.metal` softmax merge). Tie-break verified to match `mx.argmin` exactly (lowest index wins).
- `veloxquant_mlx/metal/src/h2o_evict_apply.metal` — given the winning index, compacts the surviving rows and re-rotates (NeoX-style RoPE, matching `a2ats_rope.py`'s convention) exactly the rows whose position shifted. Rows before the eviction gap are **bit-identical copies**, not just numerically close — verified explicitly.
- `veloxquant_mlx/metal/_h2o_evict.py` — Python wrapper (`h2o_fused_evict`), following this repo's existing kernel house style (`_comm_vq.py`, `_scalar_attend.py`).
- Wired into `h2o_update` behind the existing `metal_available()` capability check — pure-MLX fallback kept, per this repo's "works without Metal" policy.
- `veloxquant_mlx/tests/metal/test_h2o_evict.py` — 18 new parity tests: sink protection, interior eviction with exact RoPE recovery, tie-break-matches-`mx.argmin`, bit-identical untouched rows, determinism, large-`n_total` grid-stride stress (up to 2048 rows), multi-group independence, input validation, and a benchmark.

**Design decision:** two separate dispatches (find-argmin, then apply) rather than one barrier-fused kernel — avoids capping the largest supportable `h2o_budget` to whatever fits in one threadgroup's grid-stride reduction, at the cost of one extra dispatch per token. Two-dispatch is simpler and lower-risk; a fused single-dispatch version is a possible future follow-up if profiling ever shows dispatch overhead (not per-row compute) still dominates.

### A real crash the kernel's speed exposed (also fixed here)

`h2o_update`'s per-token loop queues one eviction's worth of graph nodes per iteration without ever forcing evaluation. The Metal kernel's dispatches apparently register more Metal-side resources per call than the pure-MLX ops they replace — so a heavy-eviction long prefill (~3238 tokens, `h2o_budget=256`) that succeeded on the pure-MLX path (233s) **crashed** on the fused-kernel path with the same `RuntimeError: [metal::malloc] Resource limit (499000) exceeded` the earlier vectorization fix (this PR, prior commit) addressed for the below-budget case.

Fixed by flushing (`mx.eval`) every 32 loop iterations, applied to both the Metal and pure-MLX eviction paths for consistency.

### Verified before/after (real `mlx_lm.generate()`, not synthetic)

| Scenario | Before (pure-MLX) | After (Metal kernel + flush) |
|---|---|---|
| Short-prompt tight-budget scenarios (from earlier in this PR) | — | **Byte-identical output** to pure-MLX |
| `budget=256`, ~3238-token prompt, 300 generated tokens | 233.44s, `"of of of of..."` | **34.90s** (6.7x faster), `"of of of of..."` — **byte-identical output** |
| Pure-MLX path with the flush added (isolated) | 233.44s | 229.57s (negligible flush overhead), byte-identical output |

This confirms the kernel + flush fix is a pure performance change — zero effect on the already-known early-token-freeze behavior tracked in #137, which remains open and unaddressed by this PR.

### Testing

- Full suite: **1809/1809 passing**.
- Ruff clean.
- 18 new kernel parity tests + 1 new regression test for the eval-flush fix.

### Not included in this PR

A detailed technical design spec was written and used to plan this kernel (threadgroup layout, the two-dispatch vs. fused-barrier trade-off, the full T1-T8 correctness test plan) but is being kept as a local working document rather than committed to the repo.

rajveer43:
## Scope extended once more: `h2o_grace` fixes the actual freeze bug (the root cause tracked in #137)

Everything up to this point in the PR (RoPE fixes, scalability fix, Metal kernel) was performance/correctness work that left the *actual* repetition/collapse bug from #137 completely untouched by design. This commit addresses that root cause directly.

### The fix

New `h2o_grace` config field (default `16`): the most-recently-arrived `grace` tokens are protected from eviction the same way sink tokens are (treated as `+inf` in the eviction argmin). This gives every new token `grace` update steps to actually accumulate attention mass before it can be evicted — closing the loophole where a brand-new token's starting score of `0.0` made it almost always the eviction target the instant the cache filled.

Implemented identically in both paths:
- `_evict_via_mlx` — extends the existing sink-protection logic with a second `+inf` mask over the trailing `grace` indices.
- `h2o_evict_reduce.metal` — same protection, added as a new `n_grace_arr` kernel input; protects `i >= n_total - n_grace` in addition to the existing `i < n_sink`.

Verified bit-for-bit equivalent between the two paths (existing 18-test Metal parity suite still passes). `h2o_grace=0` reproduces the original paper-faithful (freeze-prone) behavior exactly — kept as an explicit, documented opt-in for anyone studying Algorithm 1 as published.

### Real-model verification — and an honest limitation found

Confirmed structurally: the kept-position window, which used to freeze permanently at `[0, budget-1]` (the prompt) with `grace=0`, now genuinely advances as generation proceeds. E.g. with `grace=8, budget=14`: positions went from `[0..13]` (frozen forever) to `[0,1,2,3,4,5,27,29,31,33,35,37,39,41]` — generated tokens are finally entering and surviving in the cache.

**But this alone doesn't guarantee coherent output.** Tested across several grace/budget combinations:

| `h2o_grace` | `h2o_budget` | Kept window | Output |
|---|---|---|---|
| 0 | 14 | frozen `[0..13]` | `Paris.\nThe\nThe answer in French...` |
| 8 | 14 | gapped, advancing | `Paris.\nTheTheThe...` (still degenerates) |
| 16 | 32 | gapped | `...the largest largest largest...` (different repetition) |
| 32 | 128 | **contiguous** (enough headroom) | **Coherent** photosynthesis explanation |

The pattern: when `h2o_budget` is tight relative to `h2o_n_sink + h2o_grace`, eviction is forced into a sparse gapped window to make room for both — and local coherence still breaks down, just via a different mechanism (broken local context from gaps) than the original permanent freeze. Coherent output needs `h2o_budget` sized with real headroom, not just `h2o_grace > 0`.

Documented precisely in `docs-site/docs/algorithms/h2o.md` (new "Grace-period testing" section) rather than overclaiming the freeze fix as a complete solution.

### Testing

- Full suite: **1811/1811 passing**.
- New tests: `test_default_grace_is_nonzero`, `test_grace_protects_new_tokens_from_immediate_eviction` (cache-level); grace protection covered in the Metal kernel's existing parity suite.
- Fixture updates: several pre-existing tests using small budgets (4-16, predating `h2o_grace`) now pin `h2o_grace=0` explicitly, since they test the underlying eviction mechanism in isolation, not the grace feature — including the `CaMKVCache`/`ChunkKVCache` tests that assert bit-for-bit equivalence with H2O's core eviction rule.
- Docs build clean, ruff clean.

This is now a real, if partial, fix for the issue tracked in #137 — see that issue for the updated summary.

rajveer43:
## Scope extended once more: `h2o_decay` fixes stale-token dominance that grace alone left behind

The previous commit (`h2o_grace`) fixed the permanent-freeze bug but explicitly flagged an honest limitation: fixing the freeze wasn't enough for coherent output at tight budgets. Investigated that limitation further — it's actually two separable things, and this commit fixes one of them.

### Root cause found

Scores are a running sum that never shrinks. A token that survived its `grace` window early in generation keeps accumulating mass forever, so it can permanently outrank every token that graduates out of grace afterward — the eviction argmin then thins the *rest* of the budget instead of tracking the current generation position.

**Verified directly:** `h2o_budget=64, h2o_n_sink=4, h2o_grace=16`, 400 decode steps. Without decay, the kept window was `[4, 5, ..., 74]` — essentially still the prompt, 300+ positions behind where generation actually was. With `h2o_decay=0.98`, the same run's kept window was `[297, 299, ..., 415]` — genuinely tracking the current position.

### The fix

`h2o_decay` (default `0.98`): every existing score is multiplied by `decay` each update step, before new attention mass is added. `decay=1.0` is a byte-for-bit no-op (verified). Implemented identically in the per-token loop and the vectorized batch-absorb path (`_batch_absorb_no_eviction`) — verified numerically equivalent to fp32 rounding in both the from-empty and nonempty-prior-state branches, since the exponentiated per-row decay power was the highest-risk part of this change.

No Metal kernel changes needed — decay is applied entirely during score accumulation, before either eviction path (`_evict_via_mlx` / `_evict_via_metal`) ever runs, so the existing 18-test Metal parity suite is unaffected and still passes unmodified.

### What decay does NOT fix — tested and ruled in/out explicitly

I did not want to repeat the mistake of overclaiming a fix, so I specifically tried to break this one:

1. **The gap-of-2 pattern at very tight headroom is structural, not staleness.** At `budget=14, n_sink=2, grace=8` (only 4 "competable" slots), the non-sink kept window has a steady every-other-position gap **with or without decay** — traced it directly: the token that just graduated out of grace is always the newest competable token, so it's close to the sole minimum by construction, independent of decay strength. Documented as a distinct, still-open effect, not folded into the decay fix's claims.

2. **A/B'd decay against a case with real headroom** (`budget=128, n_sink=2, grace=32`) to isolate whether decay changes eviction *order* or just score *scale*: kept positions and generated text were **byte-for-byte identical** between `decay=1.0` and `decay=0.98`, despite score magnitudes differing ~8x (max score 24.16 vs 2.97). Decay didn't need to change anything here because grace+headroom already kept eviction rare.

3. **Found a third, separate bug while doing that A/B** — at that same `budget=128` config, generation breaks mid-word (`"...the th th th th..."` instead of `"...the thylakoid..."`), identically regardless of decay. A no-eviction control run on the same prompt/seed stays fully coherent for the entire 300-token generation, proving this is a real eviction-quality issue independent of both grace and decay — not something this commit fixes, and not silently swept under "coherent output" claims. Documented as an open item in `docs-site/docs/algorithms/h2o.md`'s new "Grace-and-decay testing" section.

### Testing

- Full suite: **1824/1824 passing**.
- New tests: validation range checks (`decay` must be in `(0, 1]`), no-op equivalence at `decay=1.0`, staleness-prevention assertions, and batch-vs-sequential decay parity (both branches) in `test_h2o.py`; cache-level default/validation/combined-grace+decay tests in `test_h2o_cache.py`.
- Fixture updates: `test_h2o_cache.py`'s `_make()` helper and the CaM/ChunkKV cross-equivalence tests now pin `h2o_decay=1.0` alongside the existing `h2o_grace=0` pin, for the same reason — isolating the base eviction mechanism from either fix.
- Docs build clean (`npm run build`), ruff format clean.

This closes the loop on the score-based issues tracked in #137. The remaining mid-generation eviction-quality gap is real but distinct — happy to open a new issue for it if useful rather than keep stretching this one's scope.

rajveer43:
Splitting this into three focused PRs for review, since it bundled a pure correctness fix, a perf-only kernel, and a behavior/default-changing algorithm fix into one review — see discussion above. Closing in favor of:

1. #139 — RoPE position desync + prefill-scalability crash (pure correctness, no behavior change)
2. #140 — fused Metal eviction kernel (pure perf, stacked on #139)
3. #141 — grace period + score decay (changes eviction behavior and two defaults; still has a known-open eviction-quality gap — needs the most scrutiny)

All commits and test coverage carry over unchanged; nothing here is lost, just regrouped for review.

Issue #148, opened 2026-08-11
CurDKV/H2O: RoPE remap uses plain math, but Llama-3-family models use scaled RoPE — remapped rows get wrong rotation


rajveer43:
## Summary

While fixing #147's follow-on (a separate bug: \`CurDKVKVCache\` reset \`self.offset\` to the post-eviction kept-row count every call instead of the true absolute step count, causing \`mlx_lm\`'s \`self.rope(x, offset=cache.offset)\` to rotate the next query/key at the wrong position — now fixed on the \`fix/curdkv-eviction-age-bias\` branch, following H2O-adapted's already-proven \`self.offset\`/\`rope_remap_positions\` pattern), real-model verification surfaced a second, deeper, **not yet fixed** issue: the RoPE re-rotation itself is wrong on Llama-3-family models.

## Root cause

\`rope_remap_positions\` (\`veloxquant_mlx/quantizers/a2ats_rope.py\`) implements plain RoPE:

\`\`\`python
inv_freq = 1.0 / (base ** (mx.arange(0, half, dtype=mx.float32) / half))
\`\`\`

This matches \`mlx_lm\`'s **default** rotation. But Llama-3-family models (confirmed on \`mlx-community/Llama-3.2-1B-Instruct-4bit\`) are configured with:

\`\`\`python
rope_scaling = {'factor': 32.0, 'high_freq_factor': 4.0, 'low_freq_factor': 1.0,
                 'original_max_position_embeddings': 8192, 'rope_type': 'llama3'}
\`\`\`

\`mlx_lm.models.rope_utils.Llama3RoPE\` implements a **piecewise, per-frequency-band rescaling** on top of the base rotation (unscaled for high frequencies, fully-scaled for low frequencies, smoothly interpolated in between — see \`initialize_rope\`/\`Llama3RoPE.__init__\` in that file). \`rope_remap_positions\`'s plain-RoPE formula does not reproduce this, **regardless of which \`base\`/\`rope_theta\` value is passed** — the mismatch is in the per-frequency scaling curve, not just the base frequency.

## Evidence

After fixing the \`self.offset\` desync (verified via a synthetic bulk-vs-split-call reproduction that now passes bit-exact), I compared CurDKV's retained key rows against an uncompressed reference cache on a real \`Llama-3.2-1B-Instruct\` generation (140-token prompt, \`curdkv_budget=138\`, so only 2 evictions occur):

[...]

PR #151, opened 2026-08-11
feat(transfer): cross-model KV cache transfer via closed-form ridge mapper (arXiv:2608.03893)

### LINKED ISSUE #150: Cross-model KV cache transfer: closed-form per-head ridge mapper for prefill reuse (NVIDIA, 2026)
## Paper

**Cross-Model KV Cache Transfer in LLM Families: A Closed-Form Linear Mapping for Prefill Reuse** — Heo, Shafipour, Zhao, Golub, Kamani, Borkar, Chandran, Zardoshti, Darvish Rouhani (NVIDIA).

## What it does

When production serving swaps between different-sized models in a family (cost-quality cascading, mid-conversation switching, routing), the receiving model normally has to re-run prefill over the whole accumulated context. This paper skips that: it maps the *source* model's KV cache into the *target* model's expected format with a closed-form, training-free linear map, and the target decodes directly from the mapped cache.

The core empirical finding is that cross-model KV has substantial linear structure. On Qwen3 14B→32B, a single source layer explains 56% of variance in the target's keys and 32% in values, rising to 79%/65% when several source layers are combined.

The mapper has three components:

1. **Per-head ridge regression** (λ=0.01), fit independently for each target (layer, head) via the closed form `W* = (XᵀX + λI)⁻¹ XᵀY`. No gradients.
2. **Cross-layer source selection** — each target layer draws from its top-`k` most predictive source layers, concatenated. This is the single largest contributor: dropping `k` from 8 to 1 takes K `R²` from 0.79 to 0.56 and collapses downstream accuracy.
3. **Content-space (RoPE-stripped) mapping** — keys are un-rotated by `R_Θ⁻¹` before the fit and re-rotated with the target's RoPE at inference, so the fit is position-free and transfers across context lengths.

Calibration is 500 FineWeb-Edu sequences of 1024 tokens, stride-4 subsampled to ~128K tokens per target head.

**Reported results:** across six matched-KV pairs in three families (Qwen3, Llama 3.1, Ministral 3), four pairs retain 73–98% of the receiver's standalone accuracy across five benchmarks; two degrade sharply. The mapper runs 2.7–25× faster than re-prefill. A nonlinear MLP recovers up to +37 pp HellaSwag retention on the failure pairs. Notably, calibration `R²` does *not* predict per-pair retention (r = −0.20) — attention-output cosine does (r = +0.57), because what matters is *where* the residual error lands relative to attention-sensitive subspaces, not its magnitude.

## Why this doesn't fit the existing `method="..."` contract

Every one of our 41 methods is a single-model cache compressor: `KVCacheBuilder.for_model(model, config)` builds per-layer caches that swap into one model's `make_cache`, compressing K/V as they are appended. Cross-model transfer is a structurally different capability:

- It needs **two models** (source + target) present, not one.
- It has an **offline fitting phase** producing a persisted artifact — the authors report ~47–87 min per pair on an 8×H100 node, and mapper artifacts of 1.01–3.36 B parameters (4–12 GB per pair).
- It **does not compress anything.** It trades prefill compute for a matmul. Our compression-ratio and `memory_bytes()` telemetry is meaningless here, and `MethodFamily` (quantization / eviction / hybrid) has no correct slot for it.

Forcing it into `KVCacheConfig.method` would require special-casing the builder, the registry probe, the serve tier logic, and the telemetry surface. So the plan is a **separate `veloxquant_mlx/transfer/` subsystem** with its own API, leaving the 41-method contract untouched.

## Apple Silicon scope

The paper's pairs (14B→32B, 8B→70B) assume datacenter memory. Holding source + target + a multi-GB mapper resident is only realistic on Apple Silicon at the small end of a family — e.g. Qwen3 0.6B→1.7B or 0.6B→4B. Two consequences:

- Mapper artifacts should be **memory-mapped from disk per layer**, not held resident in full.
- The retention numbers above are for the paper's pairs. **We should not restate them as our own** until measured on the small pairs that actually fit; small-family pairs may behave differently, and the paper itself shows matched-KV does not guarantee success (two of six pairs failed).

The `matched-KV` precondition (source and target share KV head count and per-head dimension) must be validated at fit time and refused loudly when violated — the paper only tests matched-KV pairs and explicitly leaves mismatched-KV to future work.

## Planned scope for the first PR

- `veloxquant_mlx/transfer/` subsystem: per-head ridge fit, top-`k` cross-layer source selection, RoPE-stripped content-space mapping, artifact save/load, matched-KV validation.
- A Metal kernel for the RoPE strip / re-apply path (the per-token rotations in `strip → map → re-encode` are fusable).
- Tests + docs page.

**Deferred to follow-ups:** the nonlinear MLP variant (§4.4 — needs gradient training, a new dependency shape for this library), the full five-benchmark evaluation harness, multi-turn handoff evaluation, and the attention-output-cosine diagnostic (§4.5).

Honest caveat up front: the inference-path speedup is the paper's headline, but the offline fit is expensive and the artifact is large. Whether this is *practical* on Apple Silicon at small-model scale is exactly what the first PR should let us measure, not something we should assert in advance.

rajveer43:
Closes #150.

Implements ["Cross-Model KV Cache Transfer in LLM Families: A Closed-Form Linear Mapping for Prefill Reuse"](https://arxiv.org/abs/2608.03893) (Heo et al., NVIDIA, arXiv:2608.03893) as a new `veloxquant_mlx.transfer` subsystem.

## What this is, and what it is not

**Not a compression method, and deliberately not in the 41.** Every other method here shrinks one model's cache behind `KVCacheConfig(method="...")`. This maps a *source* model's already-prefilled KV into a *target* model's format so the receiver can skip prefill when you swap models mid-session. Cache size is unchanged — what you save is prefill compute.

| | Other methods | This |
|---|---|---|
| Models | 1 | **2** |
| Reduces | cache bytes | **prefill compute** |
| Cache size after | smaller | unchanged |
| Setup | none / short calibration | **offline per-pair fit** |

It lives in its own subsystem because the single-model, per-layer cache contract cannot express a two-model mapping with an offline fit phase and a multi-GB artifact. `KVCacheConfig`'s method Literal, the registry, `MethodFamily`, and the serve-tier probe are all untouched — the README still says 41 methods, and that count is still correct.

## The mapping

```
K̂_t = (strip_rope(K_s) · W_K + b_K) · target_rope
V̂_t =            V_s    · W_V + b_V
```

1. **Top-`k` cross-layer source selection** (§3.2) — each target layer fit from the `k` source layers that best predict it. The paper's largest single contributor; `k=1` collapses accuracy.
2. **Per-head closed-form ridge** (§3.1) — `W = (XᵀX + λI)⁻¹ XᵀY` on centered data so the intercept escapes the penalty. Solved via `mx.linalg.solve` in fp32, since the Gram matrix is near-singular by construction (selected source layers are correlated *because* they were chosen for joint predictiveness). No gradient training.
3. **Content-space mapping** (§3.3) — keys un-rotated before fitting, re-rotated with the target's RoPE at apply time, so the fit is position-free.

## Metal kernel

Source and target generally have **different `rope_theta`**, so unlike `h2o_evict_apply`'s same-base remap, the strip/re-apply pair does *not* collapse to a single delta rotation. `crosskv_rope_recode` fuses both into one dispatch via the per-dimension angle difference:

```
angle = p · (θ_t^(-d/half) − θ_s^(-d/half))
```

Inverse frequencies are subtracted *before* multiplying by position — at `p` in the thousands both products are large and nearly equal, so differencing post-multiply loses significant bits.

**Measured 4.6×–9.6× over the two-pass MLX path**, ~6–9× at long context (full table and methodology in [this comment](https://github.com/rajveer43/VeloxQuant-MLX/pull/151#issuecomment-5252876997)). Output matches the reference to fp16 storage precision. Wired into `recode_rope` for 3-D inputs with an MLX fallback when Metal is unavailable.

> An earlier revision of this description claimed 12.9×. That figure was measured without warmup or `mx.synchronize()` and is **not reproducible** — corrected above.

Note this accelerates the *cheap* half: the dominant cost of the subsystem is the offline ridge fit, which the kernel does not touch.

## Artifact handling

The paper's mappers run 1.01–3.36 B params (4–12 GB) per pair. Stored as safetensors + JSON manifest, loaded **lazily per layer** and released after use, so peak memory is one layer's weights rather than the whole artifact — the reason this is usable on Apple Silicon at all. Saving a partially-unloaded mapper raises rather than silently writing an incomplete file.

## Refused rather than approximated

`MatchedKVError` on mismatched KV shapes, `rope_scaling`, and `traditional=True` RoPE. The paper only evaluates matched-KV pairs; mapping outside that regime produces plausible, wrong keys — the worst failure mode. (`rope_scaling` is the same gap tracked in #148 for the eviction caches.)

## Honesty notes

- **The paper's numbers are not ours.** 73–98% retention and 2.7–25× speedup were measured on 14B→32B and 8B→70B pairs on 8×H100 nodes. Not reproduced here. What *is* verified: the implementation recovers a known linear relationship exactly (R²=1.0, reconstruction error ~5e-4) and the kernel matches its reference.
- **Two of the paper's own six pairs failed badly** (42–44% avg retention). Matched KV correlates with success but does not guarantee it.
- **`R²` will mislead you.** Paper §4.5: calibration `R²` does *not* predict downstream retention across pairs (r = −0.20); attention-output cosine does (r = +0.57). We record `R²` per layer for source selection and debugging only, and say so.
- **Apple Silicon scale limit.** Source + target + mapper resident is realistic only at the small end of a family (e.g. Qwen3 0.6B→1.7B).

## Not implemented (documented, not silently missing)

- Nonlinear MLP variant (§4.4) — recovers up to +37 pp where ridge fails; needs gradient training, a new dependency shape for this repo.
- Attention-output cosine diagnostic (§4.5) — the metric that actually predicts retention.
- Five-benchmark evaluation harness and multi-turn drift measurement (§4.6).
- Cross-family and mismatched-KV transfer — open in the paper too.

## Testing

**+60 tests, 1945 → 2005 passing.** The 4 failures in `test_registry.py` / `test_panel*.py` are **pre-existing on master** (verified by stashing) and unrelated — they concern the serve-tier default.

Tests are built around closed-form-known answers so subtle errors fail visibly rather than merely scoring worse: exact RoPE round-trip, fused-vs-two-step equivalence, ridge recovering a planted `(W, b)`, source selection finding the planted driver layer, and a guard that output keys carry the **target's** rotation (stripping with the source base must *fail* to recover content space).

`ruff format` and `ruff check` clean on all new files. Docusaurus build passes with the new page and sidebar entry.

## Docs

- New page: `docs-site/docs/algorithms/cross-model-transfer.md`, following the repo's "honesty crux" convention.
- README: a short section positioning this as explicitly *outside* the 41-method count.
- `CITATIONS.md`: full entry with adaptation notes.


rajveer43:
## Metal kernel benchmarks — and a correction to the 12.9× figure

I re-ran the `crosskv_rope_recode` benchmark properly and **the 12.9× claim in the PR description does not hold up.** Posting the corrected numbers rather than leaving it standing.

### What was wrong

The original measurement had no warmup and no `mx.synchronize()`, so it was partly timing async dispatch rather than completed GPU work. Repeated runs of that same harness gave 12.9×, then 8.3×, then 5.7× — the spread alone shows it wasn't measuring what I claimed.

### Corrected methodology

15 warmup iterations, `mx.synchronize()` on both sides of each timing window, 25 iterations per sample, 7 samples, **median** reported (not min — min flatters the GPU path).

### Results

`veloxquant_mlx.transfer.rope` two-pass (`strip_rope` → `apply_rope`) vs. fused `crosskv_rope_recode`, Apple Silicon (arm64), MLX 0.32.0, `θ_s=1e4 → θ_t=1e6`:

| shape `[BH, N, D]` | dtype | two-pass | fused | speedup | max err |
|---|---|---|---|---|---|
| `(8, 1024, 128)` | fp16 | 1.29 ms | 0.28 ms | **4.57×** | 3.9e-03 |
| `(8, 4096, 128)` | fp16 | 5.85 ms | 1.02 ms | **5.74×** | 3.9e-03 |
| `(8, 8192, 128)` | fp16 | 9.21 ms | 0.96 ms | **9.63×** | 3.9e-03 |
| `(4, 16384, 128)` | fp16 | 12.04 ms | 1.63 ms | **7.39×** | 3.9e-03 |
| `(8, 8192, 128)` | fp32 | 7.72 ms | 1.25 ms | **6.17×** | 1.8e-03 |

**Honest summary: 4.6×–9.6× depending on shape, roughly 6–9× at long context.** Not a single headline number — it varies with sequence length and dtype, and there is real run-to-run variance on a thermally unconstrained laptop GPU. The `9.63×` cell should not be quoted in isolation any more than `12.9×` should have been.

`max err` is the elementwise deviation from the MLX reference. 3.9e-03 is one fp16 ULP at these magnitudes — i.e. the kernel agrees with the reference to storage precision, and the fp32 row confirms the gap is representation, not arithmetic.

### Why there is a speedup at all

Both rotations act on the same `(d, d + D/2)` pair at the same position, so they compose into one rotation by the per-dimension angle difference — one `sincos` and one 2×2 rotation per pair, no intermediate `[N, D]` array, no materialized cos/sin tables. The two-pass MLX path allocates and traverses the full tensor twice.

### Also worth stating plainly

This kernel accelerates **one step of the apply path**, which was already the cheap half. The dominant cost of this subsystem is the **offline ridge fit** (the paper reports ~47–87 min per pair on 8×H100), and that is untouched by this kernel. A 6–9× win on the RoPE re-encode does not make the subsystem 6–9× faster end to end.

Updating the PR description and the docs page to match these numbers.

PR #155, opened 2026-08-11
Rewrite the playground for a non-expert audience

### LINKED ISSUE #154: Playground reads like internal docs — rewrite for a non-expert audience
## Problem

The playground is written for someone who already knows what a KV cache is. That is the wrong audience — it is the page we point newcomers at from the landing hero.

Before a visitor makes a single choice, they meet: *KV-cache compression method*, *key accounting ~16×*, *full-KV*, *zero-calibration*, *eviction*, *head_dim*, *RVQ*. The **Goal** dropdown — the third control on the page — shows raw internal enum values:

```
Max key accounting (~16×)
Constant memory (eviction)
```

These are API identifiers rendered directly into the UI.

### Six concrete issues

**1. The headline makes a claim the page then walks back.**
`updateHeadline()` renders *"Fit up to 7.5× more context on your Mac."* Three sections later the same number carries *"Accounting-only: … resident RAM may not drop by the full factor."* The caveat is correct and must stay — but the headline should not have needed it. Lead with something unconditionally true.

**2. "Key-accounting ratio" is the largest number on the page.**
The hero stat row leads with an internal metric. A consumer wants to know what to run and how long their conversation can get.

**3. The three steps are named after the software, not the user's question.**
*Recommend / Explore / Verify* describes what the tool does. A newcomer is asking *"what should I use?"*, *"how much fits?"*, *"is this real?"*.

**4. Jargon is never defined at first use.**
A `.pg-tip` tooltip component already exists but is used exactly once (the resident-RAM pill).

**5. `Sequence length` is buried under "Advanced — architecture shape".**
It is arguably the most intuitive knob on the page — *how long is my conversation?* — hidden behind the scariest label, next to `head_dim`.

**6. Method blurbs read like release notes.**
*"Per-channel keys, throughput-neutral"* is not consumer copy.

## Scope

Copy, labelling and information hierarchy. **No computed value changes.** `recommend()`, `estimateKvFp16Mb`, `METHODS[].ratio` and the benchmark JSON keep identical behaviour — verified by recomputing an identical setup before and after.

### Both sides of the port

`landing/assets/playground.js` is a hand-maintained 1:1 port of `veloxquant_mlx/tools/mac_recommender.py`, including the `warnings` strings. Those get rewritten in plain English **on both sides together**, so `veloxquant recommend` gets the friendlier wording too rather than drifting from the web page.

Three things constrain that rewrite:

- `tests/non_metal/test_mac_recommender.py` asserts on substrings `"headroom"` and `"tight"` (lines 59, 64). New wording must keep those words.
- `README.md:388` quotes one warning verbatim — updated in the same commit.
- `veloxquant_mlx/cli/recommend.py` prints warnings straight through, so it inherits the new text for free.

## Honesty guarantees that must survive

Reworded into plain English, never dropped:

- the `resident_savings_likely` / accounting-only distinction
- the `~450 tokens/page` approximation label on the book comparison
- the "Measured on real hardware · source `figures/…`" provenance line
- every entry in `res.warnings` still reaching the screen

## Other constraints

- Zero-build, vanilla JS, self-hosted fonts — no new dependency, no CDN.
- `role="tablist"` / `radiogroup`, roving `tabindex`, arrow-key nav, `aria-live` and `prefers-reduced-motion` all keep working.
- `restoreFromUrl` / `buildShareUrl` still round-trip. Renamed controls keep their existing `URL_KEYS` query key, so links already shared do not break.

## Acceptance

- [ ] Same setup before/after produces identical MB, %, and token counts
- [ ] A pre-existing share URL still restores correctly
- [ ] Every `res.warnings` entry still renders (check 8 GB + 32B)
- [ ] Keyboard: stepper arrows, method cards, benchmark segmented control
- [ ] `pytest tests/non_metal/test_mac_recommender.py` green
- [ ] PR body carries a before → after table for every user-visible string, so the rewrite is reviewable as prose rather than as a diff

rajveer43:
Closes #154.

Rewrites the playground's language and information order for someone who has never heard of a KV cache. **This is a copy and hierarchy change — no computed value moves.**

Because `landing/assets/playground.js` is a hand-maintained 1:1 port of `veloxquant_mlx/tools/mac_recommender.py`, the warning and rationale strings are rewritten **on both sides together**, so `veloxquant recommend` gets the plainer wording rather than drifting from the web page.

---

## The two judgement calls

**The headline no longer promises something the page walks back.** It read *"Fit up to 7.5× more context on your Mac"* — but three sections later the same number carried *"Accounting-only: resident RAM may not drop by the full factor."* The caveat was correct; the headline should not have needed one. It now leads with the **percentage saved**, which is unconditionally true for every method.

**The negative-headroom warning was wrong, not just jargony.** Rewriting it surfaced that `32B on 8 GB` — which is 14 GB short — produced *"barely fits"*. That is a real (if cosmetic) bug: the softer phrasing reads as a caution about a workable setup rather than an impossible one. Negative headroom now says *"will not fit"* and states the shortfall. No numeric output changed; it is a second branch on the same threshold.

---

### Controls

| | Before | After |
|---|---|---|
| Goal field | `Goal` | `What matters most` |
| goal=everyday | `Everyday (zero-calibration)` | `Just work well by default` |
| goal=max_key_accounting | `Max key accounting (~16×)` | `Squeeze the cache as small as possible` |
| goal=max_context | `Max context (full-KV)` | `Fit the longest conversation` |
| goal=best_quality | `Best quality` | `Keep answers as accurate as possible` |
| goal=constant_memory | `Constant memory (eviction)` | `Never grow past a fixed memory limit` |
| Presets | `Start from a model` | `Which model do you want to run?` |
| Custom chip | `Custom` | `Something else` |
| Advanced | `Advanced — architecture shape` | `Advanced — exact model shape` (+ note: only needed if your model isn't listed) |
| Seq length | `Sequence length` *(inside Advanced)* | `How long is your conversation?` *(main bar, with a `4,096 words ≈ 9 pages` hint)* |
| Head dim | `Head dim` | `Head dimension` |
| Model class | `Model class` | `Model size` |
| Budget | `RAM budget for KV` | `Memory you can spare` |
| Budget caption | `from 16 GB Mac · auto` / `manual` | `worked out from your 16 GB Mac` / `your own figure` |
| Method group | `Method` | `Try a different approach` |
| Bench group | `Benchmark` | `Which result?` |
| Share button | `Copy my setup` | `Save my setup` |

### Header and steps

| | Before | After |
|---|---|---|
| h1 | Feel the power of VeloxQuant-MLX | Run bigger models on the Mac you already have |
| Eyebrow | Interactive · runs in your browser | Free · runs in your browser · nothing is sent anywhere |
| Headline | **Fit up to 7.5× more context on your Mac.** | **Your chatbot's notes get 87% smaller.** |
| Step 1 | Recommend — *a method for your setup* | What should I use? — *the right setting for your Mac* |
| Step 2 | Explore — *context you can fit* | How much fits? — *conversation length in your RAM* |
| Step 3 | Verify — *real measured benchmarks* | Is this real? — *benchmarks we measured* |

A collapsed **"New here? What this actually does"** primer defines *KV cache* once, up front, before any control asks the visitor to care about it.

### Result cards

| | Before | After |
|---|---|---|
| Hero stat 1 | `Key-accounting ratio` `7.5×` | `Smaller by` `87%` |
| Hero stat 2 | `KV cache saved` `87%` | `Notes take up` `512 MB → 68 MB` |
| Hero stat 3 | `Resident RAM` / Frees RAM / Accounting-only | `Frees actual RAM?` / Yes / Not on short chats |
| Badge | `Recommended` | `Use this` |
| Pill | frees resident RAM / accounting-only | gives memory back / smaller, but not always freed |
| Bands | Recommended method · Memory vs context · Reproduce · Notes & warnings | Why this one · How much space this saves · Run this yourself · Things to know |
| CTAs | Open in Compression Lab → / See the benchmark → | See how much fits → / Show me the proof → |
| Memory bar | `fp16 KV` / `compressed` / `≈87% smaller KV cache` | `normally` / `with VeloxQuant` / `≈87% less memory for the same conversation` |
| Lab stats | `Tokens @ fp16 · 8 GB` | `Words that fit in 8 GB — normally` |
| No warnings | No warnings for this configuration. | Nothing to watch out for — this setup is a comfortable fit. |

The raw `7.5×` ratio is not deleted — it moves into the **Why this one** band, where the technical detail belongs.

### Warnings and rationales (Python + JS together)

| Before | After |
|---|---|
| `7B 4-bit weights (~4.5 GB) leave little headroom on 16 GB (est. headroom -0.5 GB). Prefer a smaller model, eviction, or full-KV compression.` | `A 32B model will not fit in 8 GB. Its weights alone need ~18.0 GB, which is more than this Mac has once macOS takes its share — you are about 14.0 GB short of any headroom. Pick a smaller model.` *(negative case; the ≥0 case keeps "barely fits")* |
| `Metal kernel launch overhead can dominate on tiny models; prefer RVQ or disable Metal if tok/s drops.` | `On a model this small the GPU spends more time starting work than doing it, so compression can actually slow generation down. If speed drops, switch to turboquant_rvq or turn the Metal kernels off.` |
| `Tight RAM with a mid/large model: consider goal=max_context (rabitq) or goal=constant_memory (eviction) for long prompts.` | `RAM is tight for a model this size. For long prompts you will get more out of 'Fit the longest conversation' (rabitq), which compresses the whole cache, or 'Never grow past a fixed memory limit' (streaming_llm), which caps it outright.` |
| `head_dim=100 is not divisible by 8; VecInfer sub_dim must divide head_dim.` | `This model will not work with vecinfer: it splits each vector into groups of 8, and this model's head dimension (100) does not divide evenly by 8.` |
| `Eviction drops tokens; quality depends on the task. For importance-based eviction try method=h2o instead.` | `This works by forgetting older tokens, so the model can lose track of things said early in a long conversation. How much that hurts depends on the task. To drop the least-used tokens instead of simply the oldest, try method=h2o.` |
| `M1 with 32B: expect lower tok/s; memory fit still depends mainly on unified RAM.` | `A 32B model on an M1 will generate text more slowly than on a newer chip. Whether it fits at all, though, comes down to how much RAM you have rather than which chip it is.` |
| `Resident RSS savings are unlikely at short context for this method's default path (accounting ratio still valid).` | `This method measures smaller but may not free much actual RAM on short prompts, because its default path unpacks values back to full precision as it reads them. The size figure is real; treat it as a measure of how well the data compresses, not as RAM you get back.` |
| `Zero-calibration default. Key accounting ~7.5x at head_dim=128. Default path dequantizes into parent fp16 cache.` | `The safe everyday pick: it works out of the box with no setup step, and shrinks the key half of the cache by about 7.5x. It unpacks each value back to full precision as it is read, so this is a size measurement rather than a drop in live memory use.` |
| `Product VQ 1-bit path targets ~16x key accounting when head_dim is divisible by sub_dim=8. Needs calibration.` | `The smallest key cache on offer, around 16x. The trade is a one-time setup pass over sample data before you can use it, and the same unpack-on-read caveat as the everyday option.` |
| `SpectralQuant targets better reconstruction at moderate compression via eigenbasis rotation (calibration required).` | `Compresses less (about 5.3x) but reconstructs the cache more faithfully, so answers stay closest to the uncompressed model. Needs a one-time setup pass over sample data.` |
| `RaBitQ compresses keys and values. Better candidate for real context capacity gains than key-only RVQ accounting.` | `Compresses both halves of the cache, not just the keys, so the space it frees is real. That makes it the better choice when you want the longest possible conversation in the RAM you have.` |
| `Structural eviction keeps a fixed sink + window. Cache token count stays bounded regardless of generation length.` | `Keeps the first few tokens plus a sliding window of recent ones and discards the rest, so memory use stops growing no matter how long the conversation runs.` |
| knob note: `1-bit keys + MSE-b4 values; prefer fused Metal path when available` | `Compresses the whole cache; turn on the Metal kernels if available` |
| knob note: `Requires one-time codebook calibration` | `Needs a one-time setup pass over sample data before first use` |

### Method cards

| Method | Before | After |
|---|---|---|
| TurboQuant-RVQ | Zero-calibration everyday default. | Works out of the box. Start here. |
| VecInfer | Max key accounting; needs calibration. | Smallest of all, but needs a setup step first. |
| RaBitQ | Full-KV compression for long context. | Frees real memory. Best for long chats. |
| SpectralQuant | Best reconstruction quality. | Keeps answers closest to the original. |
| KIVI-2bit | Per-channel keys, throughput-neutral. | Compresses less, but doesn't slow you down. |

`METHODS[].label` stays technical (it is the badge) and `.config` is untouched — it is a real API snippet.

### Benchmark captions

| Before | After |
|---|---|
| `Hand-written Metal quantize kernel vs pure-MLX, by sequence length S (B=1, H=8, D=128).` | `How much faster our hand-written Apple GPU code compresses the cache, compared with doing the same work in plain MLX. Higher is better; S is the conversation length.` |
| `Full-KV memory, Falcon3-7B shape: fp16 vs RaBitQ (~5.95× smaller).` | `Actual memory used by the notes on a Falcon3-7B model, with and without RaBitQ — about 5.95× smaller. Shorter bars are better.` |
| `KIVI on <model> (M3, 16 GB): throughput retention vs key compression.` | `KIVI running <model> on an M3 with 16 GB. Blue shows how much of the original speed is kept; purple shows how much smaller the cache got.` |
| `Measured on real hardware · source figures/…` | `Measured on a real Mac, not estimated · you can check the raw numbers in figures/…` |

---

## Honesty guarantees — reworded, never dropped

- accounting-only vs frees-real-RAM distinction → *"smaller, but not always freed"* + the full explanation in the caveat
- `~450 tokens/page` label → *"rough estimate · counting ~450 words a page"*
- measured-data provenance line → kept, and now says explicitly that it is measured rather than estimated
- every `res.warnings` entry still reaches the screen
- the bottom disclaimer now leads with **"Where these numbers come from"** and states plainly which figures are calculated and which are measured

## Verification

**Parity across all 2400 `chip × RAM × model × goal × shape` combinations** — baseline extracted from `master`, compared three ways:

```
numeric mismatches   : 0     (method, ratio, resident flag, MB, warning count, knobs)
text drift py vs js  : 0     (warnings, rationales, knob notes — word for word)
```

The repo had **no JS↔Python parity check before this**; the file header only asked future editors to "keep parity". The harness is in the PR discussion below for anyone who wants to re-run it.

Also checked:

- `pytest tests/non_metal/test_mac_recommender.py` — **8 passed**. These assert on the substrings `"headroom"` and `"tight"`, both deliberately preserved in the new wording.
- `ruff format --check` and `ruff check` — clean on `mac_recommender.py` and `README.md`.
- `node --check playground.js` — clean.
- All **32** element ids referenced by the JS resolve against the HTML; no dangling `pg-hero-ratio` after removing that stat card.
- Both landing pages parse with balanced tags.
- The five `<option value=…>` goal values still match `ruleset_dict()["goals"]` exactly, so existing `?goal=` share links resolve.
- `URL_KEYS` unchanged — every renamed control kept its query key, including `?seq=` for the promoted conversation-length field.
- CLI output re-run by hand for the tight, impossible, and comfortable cases.

**Not automated:** I could not drive a real browser (no jsdom/Playwright here, and adding one to a deliberately zero-build page seemed the wrong trade). Visual layout and keyboard navigation want a human pass on the Netlify deploy preview — particularly the wider goal `<select>` at the 560px breakpoint.

## Notes for review

- `landing/index.html` is untouched; its links to the playground read generically ("Try the playground") and are still accurate.
- The `README.md` CLI transcript is regenerated from a real run. It had drifted independently of this work — it was **missing one of the two warnings** the command actually prints.
- New CSS is scoped to genuinely new markup (`.pg-primer`, `.pg-field-seq`, `.pg-adv-note`, `.pg-rationale-tech`) plus one width rule for the now-much-longer goal labels.

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,11 +382,19 @@ knobs={'bit_width_inlier': 1, 'seed': 42}
 key_accounting_ratio≈7.5x
 resident_savings_likely=False
 kv_fp16_mb≈512.0  kv_compressed_mb_est≈68.27
-rationale: Zero-calibration default. Key accounting ~7.5x at head_dim=128.
-           Default path dequantizes into parent fp16 cache.
+rationale: The safe everyday pick: it works out of the box with no setup
+           step, and shrinks the key half of the cache by about 7.5x. It
+           unpacks each value back to full precision as it is read, so this
+           is a size measurement rather than a drop in live memory use.
 warnings:
-  - Tight RAM with a mid/large model: consider goal=max_context (rabitq)
-    or goal=constant_memory (eviction) for long prompts.
+  - RAM is tight for a model this size. For long prompts you will get more
+    out of 'Fit the longest conversation' (rabitq), which compresses the
+    whole cache, or 'Never grow past a fixed memory limit' (streaming_llm),
+    which caps it outright.
+  - This method measures smaller but may not free much actual RAM on short
+    prompts, because its default path unpacks values back to full precision
+    as it reads them. The size figure is real; treat it as a measure of how
+    well the data compresses, not as RAM you get back.
 ```
 
 Goals: `everyday`, `max_key_accounting`, `max_context`, `best_quality`, `constant_memory`. Add `--json` for machine-readable output, or `--seq-len` / `--n-layers` / `--n-kv-heads` / `--head-dim` to match a specific model. Also available in the browser via the [Compression Lab](https://veloxquant-mlx.netlify.app/playground.html).
--- a/landing/assets/playground.js
+++ b/landing/assets/playground.js
@@ -39,19 +39,34 @@ function recommend(req) {
     // Match Python's float repr (e.g. "2.0", "18.0"): the reference message
     // uses `~{weight_gb} GB` where weight_gb is always a float.
     const weightStr = Number.isInteger(weightGb) ? weightGb.toFixed(1) : String(weightGb);
-    warnings.push(
-      `${req.model_class} 4-bit weights (~${weightStr} GB) leave little ` +
-      `headroom on ${req.ram_gb} GB (est. headroom ${headroomGb.toFixed(1)} GB). ` +
-      "Prefer a smaller model, eviction, or full-KV compression."
-    );
+    // Negative headroom means the weights alone overrun the machine, so say
+    // "will not fit" rather than "barely fits" — the softer phrasing would
+    // read as a caution about a setup that is in fact impossible.
+    if (headroomGb < 0) {
+      warnings.push(
+        `A ${req.model_class} model will not fit in ${req.ram_gb} GB. Its ` +
+        `weights alone need ~${weightStr} GB, which is more than this Mac ` +
+        "has once macOS takes its share — you are about " +
+        `${Math.abs(headroomGb).toFixed(1)} GB short of any headroom. Pick a smaller ` +
+        "model."
+      );
+    } else {
+      warnings.push(
+        `A ${req.model_class} model barely fits in ${req.ram_gb} GB. Its ` +
+        `weights alone take ~${weightStr} GB, leaving about ` +
+        `${headroomGb.toFixed(1)} GB of headroom for everything else. Try a ` +
+        "smaller model, or pick a goal that caps memory."
+      );
+    }
   }
 
   const kvFp16 = estimateKvFp16Mb(req.n_layers, req.n_kv_heads, req.head_dim, req.seq_len);
 
   if (req.model_class === "1B") {
     warnings.push(
-      "Metal kernel launch overhead can dominate on tiny models; " +
-      "prefer RVQ or disable Metal if tok/s drops."
+      "On a model this small the GPU spends more time starting work than " +
+      "doing it, so compression can actually slow generation down. If speed " +
+      "drops, switch to turboquant_rvq or turn the Metal kernels off."
     );
   }
 
@@ -64,12 +79,16 @@ function recommend(req) {
     ratio = 7.5;
     resident = false;
     rationale =
-      "Zero-calibration default. Key accounting ~7.5x at head_dim=128. " +
-      "Default path dequantizes into parent fp16 cache.";
+      "The safe everyday pick: it works out of the box with no setup step, " +
+      "and shrinks the key half of the cache by about 7.5x. It unpacks each " +
+      "value back to full precision as it is read, so this is a size " +
+      "measurement rather than a drop in live memory use.";
     if (tight && ["7B", "14B", "32B"].includes(req.model_class)) {
       warnings.push(
-        "Tight RAM with a mid/large model: consider goal=max_context " +
-        "(rabitq) or goal=constant_memory (eviction) for long prompts."
+        "RAM is tight for a model this size. For long prompts you will get " +
+        "more out of 'Fit the longest conversation' (rabitq), which " +
+        "compresses the whole cache, or 'Never grow past a fixed memory " +
+        "limit' (streaming_llm), which caps it outright."
       );
     }
   } else if (req.goal === "max_key_accounting") {
@@ -78,41 +97,49 @@ function recommend(req) {
       key_codebook_bits: 8, value_codebook_bits: 8,
       key_sub_dim: 8, value_sub_dim: 8,
       use_metal_kernels: null,
-      note: "Requires one-time codebook calibration",
+      note: "Needs a one-time setup pass over sample data before first use",
     };
     ratio = 16.0;
     resident = false;
     rationale =
-      "Product VQ 1-bit path targets ~16x key accounting when " +
-      "head_dim is divisible by sub_dim=8. Needs calibration.";
+      "The smallest key cache on offer, around 16x. The trade is a one-time " +
+      "setup pass over sample data before you can use it, and the same " +
+      "unpack-on-read caveat as the everyday option.";
     if (req.head_dim % 8 !== 0) {
       warnings.push(
-        `head_dim=${req.head_dim} is not divisible by 8; ` +
-        "VecInfer sub_dim must divide head_dim."
+        "This model will not work with vecinfer: it splits each vector " +
+        `into groups of 8, and this model's head dimension ` +
+        `(${req.head_dim}) does not divide evenly by 8.`
       );
     }
   } else if (req.goal === "best_quality") {
     method = "spectral";
-    knobs = { bit_width_inlier: 3, note: "Requires spectral rotation calibration" };
+    knobs = {
+      bit_width_inlier: 3,
+      note: "Needs a one-time setup pass over sample data before first use",
+    };
     ratio = 5.3;
     resident = false;
     rationale =
-      "SpectralQuant targets better reconstruction at moderate " +
-      "compression via eigenbasis rotation (calibration required).";
+      "Compresses less (about 5.3x) but reconstructs the cache more " +
+      "faithfully, so answers stay closest to the uncompressed model. " +
+      "Needs a one-time setup pass over sample data.";
   } else if (req.goal === "max_context") {
     method = "rabitq";
     ratio = 6.0;
     resident = true;
     if (tight) {
-      knobs = { note: "1-bit keys + MSE-b4 values; prefer fused Metal path when available" };
+      knobs = { note: "Compresses the whole cache; turn on the Metal kernels if available" };
       rationale =
-        "Full-KV compression is more likely to free resident memory " +
-        "than key-only accounting methods on tight RAM.";
+        "Compresses both halves of the cache, so it genuinely gives RAM " +
+        "back rather than only measuring smaller. That matters most on a " +
+        "machine as tight as this one.";
     } else {
-      knobs = { note: "Full KV compression for longer context in fixed RAM" };
+      knobs = { note: "Compresses the whole cache, for longer chats in the same memory" };
       rationale =
-        "RaBitQ compresses keys and values. Better candidate for " +
-        "real context capacity gains than key-only RVQ accounting.";
+        "Compresses both halves of the cache, not just the keys, so the " +
+        "space it frees is real. That makes it the better choice when you " +
+        "want the longest possible conversation in the RAM you have.";
     }
   } else if (req.goal === "constant_memory") {
     method = "streaming_llm";
@@ -120,11 +147,14 @@ function recommend(req) {
     ratio = 1.0;
     resident = true;
     rationale =
-      "Structural eviction keeps a fixed sink + window. Cache token " +
-      "count stays bounded regardless of generation length.";
+      "Keeps the first few tokens plus a sliding window of recent ones and " +
+      "discards the rest, so memory use stops growing no matter how long " +
+      "the conversation runs.";
     warnings.push(
-      "Eviction drops tokens; quality depends on the task. " +
-      "For importance-based eviction try method=h2o instead."
+      "This works by forgetting older tokens, so the model can lose track of " +
+      "things said early in a long conversation. How much that hurts depends " +
+      "on the task. To drop the least-used tokens instead of simply the " +
+      "oldest, try method=h2o."
     );
   } else {
     throw new Error(`Unknown goal: ${req.goal}`);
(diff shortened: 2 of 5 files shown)


rajveer43:
The parity harness used for the 2400-combination claim, in case anyone wants to re-run it or keep it as a script. It is deliberately not committed — it reaches into a JS file with `eval` and is a verification tool, not something to maintain.

**`js_parity.js`** — runs the browser `recommend()` under node by slicing `playground.js` down to the engine (everything before the presets marker) and stubbing `VQCalc` from `calc.js`:

```javascript
/* Run the browser recommend() under node and dump the same grid the Python
   baseline recorded, so JS-vs-Python parity can be diffed directly.

   playground.js is a browser script (no exports, touches document at load).
   We take only the top slice up to the MODEL_PRESETS marker — that covers
   VQCalc consumption + recommend() — and stub VQCalc from calc.js. */

const fs = require("fs");
const path = require("path");

const LANDING = "/Users/rajveerrathod/Work/personal_projects/turboquant_mac_implementation/landing/assets";

// calc.js assigns window.VQCalc; give it a window and harvest it.
const calcSrc = fs.readFileSync(path.join(LANDING, "calc.js"), "utf8");
const window = {};
eval(calcSrc);
const VQCalc = window.VQCalc;

// Slice playground.js down to just the engine (everything before the presets).
const pgSrc = fs.readFileSync(path.join(LANDING, "playground.js"), "utf8");
const cut = pgSrc.indexOf("/* ---------- Model presets");
if (cut < 0) throw new Error("marker not found");
const engine = pgSrc.slice(0, cut);
eval(engine); // defines recommend()

const RAMS = VQCalc.ALLOWED_RAM_GB;
const MODELS = Object.keys(VQCalc.MODEL_WEIGHT_GB_4BIT);
const GOALS = ["everyday", "max_key_accounting", "max_context", "best_quality", "constant_memory"];
const CHIPS = ["M1", "M2", "M3", "M4"];
const SHAPES = [[32, 8, 128, 4096], [28, 4, 256, 8192], [64, 8, 128, 1024]];

const out = {};
for (const chip of CHIPS)
  for (const ram of RAMS)
    for (const model of MODELS)
      for (const goal of GOALS)
        for (const [nl, nh, hd, sl] of SHAPES) {
          const r = recommend({
            chip, ram_gb: ram, model_class: model, goal,
            seq_len: sl, n_layers: nl, n_kv_heads: nh, head_dim: hd,
          });
          const knobs = {};
          for (const [k, v] of Object.entries(r.knobs)) if (true) knobs[k] = v;
          out[`${chip}|${ram}|${model}|${goal}|${nl}-${nh}-${hd}-${sl}`] = {
            method: r.method,
            ratio: r.key_accounting_ratio,
            resident: r.resident_savings_likely,
            fp16_mb: r.kv_fp16_mb,
            comp_mb: r.kv_compressed_mb_estimate,
            n_warnings: r.warnings.length,
            knobs,
            // Full text too, so JS-vs-Python wording drift is catchable.
            warnings: r.warnings,
            rationale: r.rationale,
          };
        }

fs.writeFileSync(process.argv[2], JSON.stringify(out, null, 0));
console.log(`${Object.keys(out).length} combinations recorded`);
```

**`baseline.py`** — same grid on the Python side:

```python
"""Snapshot every numeric output of recommend() across the full input grid.

Copy strings are excluded on purpose: the rewrite changes them deliberately.
What must not move is the math, so this records only the numbers plus the
*count* of warnings (a dropped warning is a regression even if reworded).
"""

import importlib.util
import json
import sys
from pathlib import Path

root = Path(__file__).resolve().parents[0]
mod_path = sys.argv[2]
spec = importlib.util.spec_from_file_location("mac_recommender", mod_path)
m = importlib.util.module_from_spec(spec)
sys.modules["mac_recommender"] = m
spec.loader.exec_module(m)

RAMS = m.ALLOWED_RAM_GB
MODELS = list(m.MODEL_WEIGHT_GB_4BIT)
GOALS = m.ruleset_dict()["goals"]
CHIPS = ["M1", "M2", "M3", "M4"]
SHAPES = [(32, 8, 128, 4096), (28, 4, 256, 8192), (64, 8, 128, 1024)]

out = {}
for chip in CHIPS:
    for ram in RAMS:
        for model in MODELS:
            for goal in GOALS:
                for (nl, nh, hd, sl) in SHAPES:
                    req = m.RecommendRequest(
                        chip=chip, ram_gb=ram, model_class=model, goal=goal,
                        seq_len=sl, n_layers=nl, n_kv_heads=nh, head_dim=hd,
                    )
                    r = m.recommend(req)
                    key = f"{chip}|{ram}|{model}|{goal}|{nl}-{nh}-{hd}-{sl}"
                    out[key] = {
                        "method": r.method,
                        "ratio": r.key_accounting_ratio,
                        "resident": r.resident_savings_likely,
                        "fp16_mb": r.kv_fp16_mb,
                        "comp_mb": r.kv_compressed_mb_estimate,
                        "n_warnings": len(r.warnings),
                        "knobs": dict(r.knobs),
                    }

Path(sys.argv[1]).write_text(json.dumps(out, indent=0, sort_keys=True))
print(f"{len(out)} combinations recorded")
```

Run as:

```bash
git show master:veloxquant_mlx/tools/mac_recommender.py > orig_recommender.py
python3 baseline.py before.json orig_recommender.py   # pre-change numbers
node js_parity.js js_after.json                       # current JS
python3 compare.py                                    # three-way diff
```

Issue #164, opened 2026-08-12
KIVI: Metal kernel for group quantization + packed KV storage (perf follow-up to #162)


rajveer43:
Follow-up to #162 / #163, which fixed KIVI's correctness. This issue covers the *performance* side, deliberately kept separate.

## Current state

`KIVIKVCache._quant_dequant_along` is pure MLX: `moveaxis` -> reshape -> `min`/`max` -> `round`/`clip` -> reconstruct -> `moveaxis` back. On `Llama-3.2-3B-Instruct-4bit` (Apple M4, 2239-token prompt) KIVI-2bit runs at ~101% of fp16 throughput, so the arithmetic is not currently a bottleneck on a memory-bound decode path.

## Why a kernel is not obviously a win yet

The round-trip **dequantizes back to fp16 before SDPA** — the cache stores a contiguous fp16 tensor, so the 2-bit codes never actually exist in memory. Compression today is an *accounting* result, not a storage result. A quantize/dequantize kernel would only fuse arithmetic that is already cheap.

The real win needs **packed storage**: keep `uint8`-packed codes plus per-group `(scale, zero)` in the cache and dequantize inside the attention kernel. That reduces the peak fp16 working set, which is the thing the docs currently (correctly) disclaim as *not* reduced.

## Suggested scope

1. `kivi_group_quant.metal` — fused per-channel/per-token group min/max + quantize + dequantize, following the existing `scalar_quantize.metal` / `_scalar_quant.py` pattern (dispatch cache, Python fallback, parity test against the MLX path).
2. Reuse `bit_pack.metal` / `bit_unpack.metal` for packed `b`-bit storage.
3. Only then consider a fused dequant-in-SDPA path, in line with `fused_sdpa.py`.

Worth benchmarking step 1 in isolation first — if it does not move end-to-end throughput, steps 2-3 are where the actual memory win lives, and step 1 alone may not be worth the maintenance surface.

## Acceptance

- Parity test: kernel output bit-identical (or within fp16 rounding) to `_quant_dequant_along`.
- Benchmark showing end-to-end throughput and **peak memory** vs the current MLX path.
- No regression in `veloxquant_mlx/tests/cache/test_kivi_cache.py`.

rajveer43:
Step 1 (fused group-quant kernel) is implemented in #166.

**Result: the kernel works and is bit-exact, but it does not move end-to-end throughput — and this issue's own framing was right about why.**

[...]

PR #166, opened 2026-08-12
feat(kivi): fused Metal kernel for asymmetric group quantization (#164 step 1)

rajveer43:
Implements **step 1** of #164 — the fused group-quantize/dequantize kernel. Steps 2–3 (packed storage, dequant-in-SDPA) are deliberately **not** attempted here; see "What this does not do".

## What landed

`kivi_group_quant_{channel,token}.metal` + `_kivi_quant.py` replace the whole `_quant_dequant_along` round-trip — `moveaxis → pad → reshape → min/max → round/clip → reconstruct → moveaxis` — with one Metal dispatch, following the existing `scalar_quantize.metal` / `_scalar_quant.py` pattern.

**Two kernels, because KIVI's two schemes have opposite memory layouts** in a `[..., S, D]` row-contiguous cache:

| scheme | groups along | mapping | reduction |
|---|---|---|---|
| values (`axis=-1`) | `D` — contiguous | SIMD group per quantization group | `simd_shuffle_xor` butterfly, no barriers |
| keys (`axis=-2`) | `S` — strided | one thread owns a whole group; consecutive threads take consecutive channels | **none needed** |

The tempting alternative for keys — transpose so the token axis is contiguous, then reuse one shader for both — is what this replaces. That transpose is a full-size materializing copy and costs more traffic than the arithmetic being fused. Killing it is where most of the speedup comes from.

`head_dim` is a compile-time constant (fixed per cache, so index math becomes shifts/masks); the sequence length stays runtime via MLX's `x_shape`.

Wired into `KIVIKVCache._quant_dequant_along`, so `SinkProtectedKVCache` inherits it automatically. `use_metal_kernels` follows `VecInferKVCache`'s three-state convention (`None` auto-detect / `True` require / `False` force-off), with a latching fallback to MLX on any kernel-side failure.

## Bit-exact, not just close

KIVI is the repo's reference baseline, so an ULP-level shift would silently move every published KIVI number and make results depend on whether Metal happened to be enabled. Three details are required, each with its own test:

| Hazard | Wrong choice | Consequence |
|---|---|---|
| Ragged final group | zero-pad | changes that group's min/max, shifts every reconstruction in it |
| Rounding | `metal::round` (half-away-from-zero) | disagrees with `mx.round` on exact `.5` codes |
| Reconstruction | `fma(q, scale, gmin)` | 1 ULP differences on ~0.02% of elements |

The `fma` one is the subtle one — it's the *more accurate* answer, but parity with the reference is the contract, so the un-fused form is forced.

Verified bit-exact over b ∈ {1,2,4,8} × g ∈ {4,16,32,64,128} × fp16/fp32 × both axes × ragged shapes, plus degenerate (`min == max`), all-zero, and huge-dynamic-range inputs — 300/300 configurations, re-run unchanged after the kernels were rewritten.

## A bug worth calling out

My first version baked the array shape into the kernel header as `#define`s. That compiles **a fresh shader per sequence length** — one per decode step — so the dispatch cache became an unbounded compilation leak and generation hung outright.

Fixed by keeping the sequence length a runtime value. `test_dispatch_cache_does_not_grow_with_sequence_length` asserts 55 distinct sequence lengths still map to exactly **one compiled variant per axis**, so this can't silently return. (`head_dim` *is* specialized on purpose — it's fixed per cache, so it contributes a constant, not growth.)

## Measured results (Apple M4)

Isolated op, `B=1 H=8 D=128 b=2 g=32`:

| flush size | keys (per-channel) | values (per-token) |
|---|---|---|
| 32 | 1.32× | 1.51× |
| 128 | 1.86× | 1.50× |
| 512 | 2.03× | 1.66× |
| 2048 | 4.42× | 3.52× |

The gain widens with flush size, so it lands on **prefill**, which quantizes the whole aged-out prompt in one group-aligned flush. Decode flushes only `group_size` tokens at a time.

Two optimizations were tried and **rejected on measurement** (documented in the kernels so they aren't re-attempted): caching the group in registers is neutral for values (1.00×) and actively *harmful* for keys (0.76× — register pressure costs more than the re-read, which hits cache); threadgroup width is noise-level at every size tested.

End-to-end, `Llama-3.2-3B-Instruct-4bit`, 1541-token prompt, 120 decode tokens:

| config | tok/s | vs fp16 | peak MB | KV comp |
|---|---|---|---|---|
| fp16 (no wrapper) | 20.32 | 100% | 2394.3 | — |
| KIVI-2bit, kernel **off** | 20.62 | 102% | 2568.2 | 4.60× |
| KIVI-2bit, kernel **on** | 20.71 | 103% | 2469.3 | 4.60× |

Generated text is **identical** with the kernel on and off.

## The honest read: ~0.5% end-to-end, and that's the ceiling

The op is up to 3.4× faster but end-to-end decode gains only ~0.5%. That is not a disappointing kernel — it's Amdahl's law, and #164 predicted it.

Measured directly: one flush costs ~14.2 ms across 28 layers, and a 120-token generation triggers 3 flushes (one per 32-token group) ≈ **42.5 ms of ~5800 ms wall time — about 0.7% of decode.** So *even an infinitely fast kernel* caps the end-to-end win at ~0.7%. At +0.5% this is already near that ceiling; there is no headroom left to chase in step 1.

**Benchmarking note:** two naive timing setups each produced a *different wrong answer* during development — one reported the kernel 3× **slower**. Repeating a single identical input lets MLX reuse work the kernel path can't, and timing A-then-B lets thermal drift land unevenly. The committed benchmark rotates a pool of distinct inputs and interleaves A/B samples, reporting medians. Worth knowing before trusting any follow-up numbers here.

## What this does not do

**Peak memory is unchanged** — this is still an accounting result, not a storage one. The round-trip dequantizes to fp16 before SDPA, so the 2-bit codes never exist in memory; the kernel only fuses arithmetic that was already cheap. The real win #164 describes needs **packed storage** (`bit_pack.metal` reuse) plus dequant-in-SDPA, which is where the memory reduction actually lives.

Given the 0.7% ceiling measured above, I'd treat steps 2–3 as the substantive work and this as the prerequisite that makes them safe to build on.

## Acceptance criteria (#164)

- [x] Parity test: bit-identical to `_quant_dequant_along`
- [x] Benchmark: end-to-end throughput **and** peak memory vs the MLX path
- [x] No regression in `tests/cache/test_kivi_cache.py`

Full suite: **2075 passed** (2020 baseline + 55 new). Lint unchanged at the repo's 648 pre-existing errors; no new ones.

Refs #164


rajveer43:
## Real-model validation: prefill vs decode, three models

Measured **prefill and decode separately** rather than blending them into one tok/s, because the two phases hit the kernel in completely different regimes:

| phase | flush shape | how often | kernel regime |
|---|---|---|---|
| **prefill** | mlx_lm chunks prompts at 2048 tokens, so flushes are `S ≤ 2048` per layer | once per chunk | large-`S` — kernel is 5.65× |
| **decode** | exactly `group_size` = 32 tokens per layer | once every 32 steps | small-`S` — kernel is 1.40× |

### Per-flush cost (Apple M4, H=8 D=128, keys+values, ×28 layers)

| flush size | kernel off | kernel on | speedup | saved per flush |
|---|---|---|---|---|
| 32 (decode) | 0.5050 ms | 0.3601 ms | **1.40×** | 4.1 ms |
| 2048 (prefill chunk) | 3.8337 ms | 0.6785 ms | **5.65×** | 88.3 ms |

Which predicts the end-to-end effect *before* measuring it:

- **prefill**, 8k prompt: 4 chunks × 88 ms = 353 ms saved of ~21 000 ms wall → **+1.7%**
- **decode**, 240 tokens: 7 flushes × 4.1 ms = 28 ms saved of ~10 000 ms wall → **+0.28%**

### Measured end-to-end (kernel on vs off)

| model | layers | KV heads | prefill | decode | identical text | KV comp |
|---|---|---|---|---|---|---|
| Llama-3.2-3B-4bit | 28 | 8 | 1.019× (2k) / 0.997× (8k) | 0.989× | ✅ | 5.02× |
| Qwen2.5-7B-4bit | 28 | **4** | 1.023× | 0.994× | ✅ | 4.94× |
| Mistral-7B-v0.3-4bit | 32 | 8 | 1.001× | 0.996× | ✅ | 4.75× |

**Every number is within ±2%, i.e. indistinguishable from noise** — the run-to-run spread on prefill alone was 281–352 tok/s (±25%) for an unchanged fp16 baseline. The predicted +1.7% / +0.3% is simply below the measurement floor of a real generation.

That is the honest headline: **the kernel is a large op-level win and an invisible end-to-end one.** Both facts are real and they are consistent — quantization is ~1–2% of prefill and ~0.3% of decode, so there was never room for more.

Qwen2.5-7B is the useful correctness case here: 4 KV heads instead of 8 gives a different flush tensor shape, and it still produces byte-identical output and the same compression ratio.

### Peak memory: no effect (and a trap I nearly reported)

| model | fp16 | kernel off | kernel on |
|---|---|---|---|
| Llama-3.2-3B | 2.604 GB | 2.726 GB | 2.599 GB |
| Qwen2.5-7B | 5.105 GB | 5.080 GB | 5.130 GB |
| Mistral-7B | 4.891 GB | 4.921 GB | 4.941 GB |

Llama alone looks like the kernel *reduces* peak memory by 127 MB — a tidy story about eliminating MLX's intermediates. Qwen and Mistral move the other way by 50 MB and 20 MB. **The direction isn't consistent, so this is allocator noise, not a benefit.** Reporting the Llama number by itself would have been wrong.

Peak memory is unchanged because the round-trip still dequantizes to fp16 before SDPA — the 2-bit codes never exist in memory. That remains the case this PR explicitly does *not* fix.

### Measurement notes

Three separate methodology traps produced wrong answers during this work, all corrected:

1. **Sync-based attribution is meaningless under lazy evaluation.** Wrapping the quantize call in `mx.synchronize()` + `mx.eval()` attributes the entire pending graph — attention, MLP, everything — to quantization. It reported quantization as *97.8% of prefill*. It is ~1–2%.
2. **Never benchmark against a busy GPU.** One microbenchmark run overlapped a background LLM job and reported 0.09× and 28.08× for the same kernels that measure 1.40× and 5.65× on an idle machine.
3. **Repeating one identical input** lets MLX reuse work the kernel path can't, which inverted an earlier comparison entirely.

The committed benchmark rotates a pool of distinct inputs, interleaves A/B samples, and reports medians.

### Bottom line

Ready to merge on its own terms: bit-exact, 1.4×–5.65× on the op, no regression on any model, identical generated text everywhere. But it will not move a throughput chart, and the PR body says so.

The end-to-end win #164 is actually after lives in **steps 2–3** — packed `uint8` storage plus dequant-in-SDPA — which is also the only path to reducing peak memory.

PR #170, opened 2026-08-14
feat(qfilters): paper-faithful query-SVD calibration + fused Metal eviction kernels

### LINKED ISSUE #169: Q-Filters: implement paper-faithful query-SVD calibration + fused Metal eviction kernels
## Summary

Q-Filters is implemented, but **not as per the paper** — and the code says so itself. `quantizers/qfilters.py` labels the gap "THE HONESTY CRUX": the paper (arXiv:2503.02812) derives the per-head filter from the **SVD of query vectors** collected offline (Eq. 1, §3.2), while this implementation derives it from the **SVD of observed keys**.

That is a different estimator of a different matrix, and it has a measured consequence already recorded in `docs-site/docs/algorithms/qfilters.md`: the key-SVD recovers the dominant axis (`filter_cosine ≈ 0.97`) but **not which end of it is important**. So `qfilters_sign` flips arbitrarily from row to row.

The paper resolves that sign for free. Theorem 3.3 gives `E⟨Qᵢ,Kⱼ⟩ ≈ κʰ⟨Kⱼ,uʰ⟩` with **`κʰ = E⟨Qᵢ,uʰ⟩ > 0`** — the positivity of `κʰ` is what fixes the direction, and it only holds when `uʰ` comes from the *query* distribution. Estimating from keys throws away exactly the quantity that disambiguates the sign.

## Gaps

1. **Estimator is key-SVD, not query-SVD** (paper §3.2 / Eq. 1). Root cause of the sign ambiguity.
2. **No Metal kernel.** Sibling evictors have fused kernels (`h2o_evict_reduce.metal`, `keyformer_evict_reduce.metal`); Q-Filters has none — scoring and top-k run as generic MLX ops.
3. **Per-head Python loop.** `cache/qfilters_cache.py` runs a nested `for b: for h:` with per-head `mx.stack` on every `update_and_fetch`.
4. **No real-model validation.** Docs state plainly: "No model-level benchmark has been run."

## Plan (this issue)

- [ ] `quantizers/qfilters_calibration.py` — offline **query**-SVD calibration, following the established `amc_calibration.py` module pattern. Query capture is already proven feasible in this repo by `a2ats_query_second_moment` (`H = E[qᵀq]`, `quantizers/a2ats.py:61`).
- [ ] Per-head filters `v₁⁺ = sgn(1ᵀu₁)·v₁` (paper §3.2 step 1c), GQA group-averaged (paper §3.2), saved/loaded as an artifact.
- [ ] Cache loads calibrated filters when present; key-SVD stays as a **documented fallback**, not the default.
- [ ] `metal/src/qfilters_score.metal` + `metal/src/qfilters_evict_apply.metal` + `metal/_qfilters_evict.py`, following the two-dispatch h2o/keyformer pattern.
- [ ] Tests: calibration correctness, sign recovery, kernel-vs-MLX parity.
- [ ] Update `docs-site/docs/algorithms/qfilters.md` — the honesty-crux section becomes accurate rather than a standing caveat.

## Deliberately out of scope

Vectorizing the B×H loop, the chunked-prefill/decode divergence, and real-model benchmarking are each tracked separately so this PR stays reviewable.

rajveer43:
Closes #169.

## The problem

Q-Filters was implemented, but **not as the paper specifies** — the code said so itself, labelling the gap "THE HONESTY CRUX". The paper (arXiv:2503.02812 §3.2, Eq. 1) derives the per-head filter from the **SVD of query activations**; the implementation derived it from the **SVD of observed keys**.

That deviation costs the **sign**. Theorem 3.3 gives

```
E<Q_i, K_j> ≈ κʰ·<K_j, uʰ>,   κʰ = E<Q_i, uʰ> > 0   (Appendix A)
```

`κʰ > 0` is what makes "higher projection ⇒ more attention" a valid ranking rule — and it is a property of the **query** distribution. Estimating from keys throws away exactly the quantity that fixes the orientation, which is why `qfilters_sign` was flipping arbitrarily from row to row.

### 1. `quantizers/qfilters_calibration.py` — the paper's actual mechanism

Follows the established `amc_calibration.py` module pattern; query capture was already proven viable here by `a2ats_query_second_moment`.

- `compute_qfilters` — per-head SVD, sign-fixed `v₁⁺ = sgn(1ᵀu₁)·v₁` (Eq. 1, §3.2 1c)
- **Deliberately not mean-centered.** Observation 3.1 is about the query cloud's *drift away from the origin*; centering subtracts that signal and returns the top *variance* axis instead — a different direction. A test pins this both ways.
- `average_gqa_filters` — GQA group-averaging (§3.2), renormalized so projections stay comparable across heads
- `collect_query_activations` — hooks `q_proj`, restores originals in a `finally` so a failed forward pass can't leave the model mutated
- `save_qfilters` / `load_qfilters` — `.npz` artifacts with version and **model-id guards**, since filters are model-specific and a silent mismatch degrades quality with no visible error

### 2. Metal kernels

`metal/_qfilters_evict.py` + `qfilters_score.metal` + `qfilters_evict_apply.metal`, following the h2o/keyformer two-dispatch pattern.

The structural difference: h2o/keyformer evict exactly **one row per token**, so dispatch 1 is an argmin and dispatch 2 a closed-form index shift. Q-Filters evicts a **whole block at once**, so there is no closed form — dispatch 1 emits the full score array and dispatch 2 compacts survivors via a cooperative scan. The threshold itself stays on the MLX side (`mx.sort`), since a top-k is a primitive MLX already implements well.

**The tie trap.** Thresholding with `score >= thresh` keeps *more* than `budget` rows whenever the threshold value repeats — and duplicates are the norm here, because every protected row shares `+inf`. Overflowing would silently break the cache's size guarantee. Rows are admitted in two tiers (strictly-greater always; equal-to-threshold only while quota remains, lowest index first), reproducing `mx.argsort`'s tie-break. Two dedicated tests cover it.

### 3. Cache wiring

`QFiltersKVCache(config, filters=...)` accepts calibrated filters. The key-SVD path is **kept as a documented fallback**, unchanged.

## Measured results

| | Calibrated (query-SVD) | Key-SVD fallback |
|---|---|---|
| Recovers planted direction **with sign** | cos > 0.99 | — |
| On pure-drift key geometry | cos > 0.99 | **< 0.3** |
| Warm-up window | none, evicts from token 0 | budget exceeded until calibrated |
| Prefill vs decode | **bit-for-bit identical** | may diverge |

The fallback's weakness turned out to be worse than "sign ambiguous": `estimate_filter_dir` mean-centers, so on drift-only geometry (the paper's `ε = −1` case) it has no signal left and returns an essentially unrelated direction.

## Testing

- **+20** calibration tests, **+20** Metal parity/guard tests, **+5** calibrated-cache tests
- Metal tests are parity checks against a numpy reference for keys *and* values across four shape/protection configurations — a kernel that is merely "reasonable" but disagrees with the reference is a silent correctness bug
- **Full suite: 2198 passed**, no regressions; `ruff check`/`format` clean on all touched files (repo-wide lint count goes 655 → 654)

## Explicitly not in scope

Tracked separately to keep this reviewable: vectorizing the per-head `B×H` Python loop in `qfilters_cache.py`, the chunked-prefill/decode divergence, and real-model benchmarking.

**The "no model-level benchmark has been run" caveat still stands** and the docs now state it for the calibration path too: this is validated on constructed query geometry, not by measuring the paper's anisotropy claim on real trained weights. The paper's accuracy/TTFT figures remain the paper's, not reproduced here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,38 @@ All notable changes to **VeloxQuant-MLX** are documented here.
 
 ### Fixed
 
+**RoPE positions after eviction in SnapKV-adapted and StreamingLLM-adapted**
+([#171](https://github.com/rajveer43/VeloxQuant-MLX/issues/171)) — both
+carried the same defect fixed earlier for Q-Filters: `self.offset` reported
+the **retained row count** rather than the true absolute token position.
+`mlx_lm` rotates both the query and the incoming key at
+`offset=cache.offset` *before* `update_and_fetch` runs, so once eviction
+started every subsequent token was rotated at the wrong position. Measured
+over 200 decode steps after a 64-token prefill at budget 32:
+
+- **StreamingLLM** — offset froze at **32** while the true position reached
+  **263** (drift **+231**, growing without bound once the window saturated).
+- **SnapKV** — offset advanced but stayed **exactly 32 behind**, the constant
+  deficit being the tokens dropped during prefill compression.
+
+Both preserve original positions (`snap_select_indices` returns kept indices
+sorted ascending; StreamingLLM's window drops rows without renumbering) and
+RoPE is relative, so reporting the true position is sufficient — survivors
+need no re-rotation, unlike H2O/Keyformer which renumber.
+
+SnapKV needed more than the `_true_offset` counter that sufficed for
+StreamingLLM: its decode path appends deltas, so the base class's cursor
+arithmetic and return slice (`self.keys[..., :self.offset, :]`) genuinely
+require the row count. `offset` is now a property yielding the row count
+while the base class is on the stack and the true position outside it.
+
+Note this diverges from the StreamingLLM paper, which assigns positions by
+index *within* the cache; matching that would require re-rotating every
+survivor each step. Recorded in the cache docstring.
+
+Two existing tests asserted `offset == retained rows` — the old meaning —
+and were updated, with a dedicated regression test added.
+
 **A2ATS-adapted paper-fidelity fixes** ([#29](https://github.com/rajveer43/VeloxQuant-MLX/issues/29)) —
 four deviations from the source paper (He et al., ACL 2025 Findings), three
 of which contradicted its equations rather than knowingly adapting them:
--- a/benchmark_scripts/qfilters_real_model_anisotropy.py
+++ b/benchmark_scripts/qfilters_real_model_anisotropy.py
@@ -1,0 +1,43 @@
+"""Test the paper's Observations 3.1/3.2 on REAL trained weights."""
+
+import sys
+
+import mlx_lm
+import numpy as np
+
+from veloxquant_mlx.quantizers.qfilters_calibration import (
+    collect_query_activations,
+    compute_qfilters,
+)
+
+mid = sys.argv[1]
+m, t = mlx_lm.load(mid)
+texts = [
+    "The capital of France is Paris, a city known for its art, cuisine and long history. " * 40,
+    "Machine learning models require large amounts of data and compute to train effectively. " * 40,
+    "In 1969, humans first walked on the surface of the moon during the Apollo 11 mission. " * 40,
+    "Photosynthesis converts light energy into chemical energy stored in glucose molecules. " * 40,
+]
+acts = collect_query_activations(m, t, texts, max_length=512, max_samples_per_head=3000)
+print(f"\n=== {mid} : {len(acts)} layers, Q shape {tuple(acts[0].shape)} ===")
+
+# Obs 3.1: E<Q_i, u^h> > 0 for the top direction.
+# Obs 3.2: projections on all OTHER SVD components have ~zero mean.
+r1, r2, energy = [], [], []
+for a in acts:
+    q = np.array(a, dtype=np.float64)  # [H, N, D]
+    f = np.array(compute_qfilters(a), dtype=np.float64)  # [H, D]
+    for h in range(q.shape[0]):
+        X = q[h]
+        proj1 = X @ f[h]
+        r1.append(proj1.mean())
+        # full SVD basis to check the OTHER components
+        _, S, Vt = np.linalg.svd(X, full_matrices=False)
+        means = np.abs(X @ Vt.T).mean(axis=0)  # |E<Q,v_m>| per component
+        r2.append(means[0] / (means[1:].mean() + 1e-12))
+        energy.append((S[0] ** 2) / (S**2).sum())
+
+r1, r2, energy = np.array(r1), np.array(r2), np.array(energy)
+print(f"Obs 3.1  E<Q,u> > 0 : {100 * np.mean(r1 > 0):.1f}% of heads   (median {np.median(r1):.3f})")
+print(f"Obs 3.2  |E<Q,v1>| / mean|E<Q,vm>|, m>=2 : median {np.median(r2):.1f}x")
+print(f"Top-component energy share : median {100 * np.median(energy):.1f}%")
--- a/benchmark_scripts/qfilters_real_model_attention_corr.py
+++ b/benchmark_scripts/qfilters_real_model_attention_corr.py
@@ -1,0 +1,95 @@
+"""Does the calibrated Q-Filter predict REAL attention? (paper Figure 4)
+
+Spearman correlation between each scorer and the observed mean attention
+S^h_t = (1/(L-t+1)) sum_{i>=t} A^h_it, measured on true attention maps from a
+real model. Compares query-SVD (calibrated), key-SVD (fallback), and K-norm.
+"""
+
+import sys
+
+import mlx.core as mx
+import mlx_lm
+import numpy as np
+from scipy.stats import spearmanr
+
+from veloxquant_mlx.quantizers.qfilters import estimate_filter_dir
+from veloxquant_mlx.quantizers.qfilters_calibration import (
+    average_gqa_filters,
+    collect_query_activations,
+    compute_qfilters,
+)
+
+mid = sys.argv[1]
+m, tok = mlx_lm.load(mid)
+inner = getattr(m, "model", m)
+args = m.args
+H, KV, D = args.num_attention_heads, args.num_key_value_heads, args.head_dim
+
+calib = [
+    "The capital of France is Paris, a city of art and history. " * 40,
+    "Machine learning models require large amounts of data to train. " * 40,
+    "In 1969, humans first walked on the surface of the moon. " * 40,
+    "Photosynthesis converts light energy into chemical energy. " * 40,
+]
+acts = collect_query_activations(m, tok, calib, max_length=512, max_samples_per_head=3000)
+qfilt = [np.array(average_gqa_filters(compute_qfilters(a), KV), dtype=np.float64) for a in acts]
+
+# --- capture K and true attention on a HELD-OUT text ---
+held = (
+    "Artificial intelligence has transformed many industries over the past decade. "
+    "Researchers continue to explore new architectures for language understanding. "
+    "The transformer architecture introduced attention as its core mechanism. "
+) * 12
+ids = tok.encode(held)[:512]
+caps = {}
+orig = []
+for li, layer in enumerate(inner.layers):
+    a = layer.self_attn
+    orig.append((a, a.k_proj, a.q_proj))
+
+    def mk(inner_fn, store, key):
+        def f(x):
+            o = inner_fn(x)
+            store[key] = np.array(o.astype(mx.float32), copy=True)
+            return o
+
+        return f
+
+    a.k_proj = mk(a.k_proj, caps, ("k", li))
+    a.q_proj = mk(a.q_proj, caps, ("q", li))
+try:
+    mx.eval(m(mx.array([ids])))
+finally:
+    for a, k, q in orig:
+        a.k_proj, a.q_proj = k, q
+
+L = len(ids)
+rows = {"query-SVD (calibrated)": [], "key-SVD (fallback)": [], "K-norm": []}
+for li in range(len(inner.layers)):
+    K = caps[("k", li)].reshape(L, KV, D).transpose(1, 0, 2)  # [KV, L, D]
+    Q = caps[("q", li)].reshape(L, H, D).transpose(1, 0, 2)  # [H,  L, D]
+    for kh in range(KV):
+        grp = Q[kh * (H // KV) : (kh + 1) * (H // KV)]  # queries sharing this KV head
+        k = K[kh].astype(np.float64)
+        # true causal attention, averaged over the group
+        S = np.zeros(L)
+        for qh in grp:
+            logits = (qh.astype(np.float64) @ k.T) / np.sqrt(D)
+            logits[np.triu_indices(L, 1)] = -np.inf
+            A = np.exp(logits - logits.max(1, keepdims=True))
+            A /= A.sum(1, keepdims=True)
+            S += np.array([A[t:, t].mean() for t in range(L)])
+        S /= len(grp)
+        kd = np.array(estimate_filter_dir(mx.array(k.astype(np.float32))), dtype=np.float64)
+        rows["query-SVD (calibrated)"].append(spearmanr(k @ qfilt[li][kh], S).statistic)
+        rows["key-SVD (fallback)"].append(spearmanr(k @ kd, S).statistic)
+        rows["K-norm"].append(spearmanr(-np.linalg.norm(k, axis=1), S).statistic)
+
+print(
+    f"\n=== {mid} — Spearman vs TRUE attention (paper Fig. 4), {len(rows['K-norm'])} KV heads ==="
+)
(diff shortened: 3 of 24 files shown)


rajveer43:
## Real-model validation added (commit 3fa0c60)

Ran the calibration path against real MLX weights. `collect_query_activations` had never executed before — it works on Llama-3.2-1B/3B and Qwen2.5-7B (`model.layers` / `self_attn.q_proj` / `args.num_attention_heads` all resolve; GQA 32→8 exercises group-averaging).

### The paper's anisotropy holds on trained weights

| Model | `E⟨Q,uʰ⟩ > 0` | Obs 3.2 ratio | Top-component energy |
|---|---|---|---|
| Llama-3.2-1B | **100%** of 512 heads | 44.4× | 90.6% |
| Llama-3.2-3B | **100%** of 672 heads | 52.3× | 84.7% |
| Qwen2.5-7B | **100%** of 784 heads | 47.9× | 81.3% |

Observation 3.1 holds in **every head measured** — that universal positivity *is* Theorem 3.3's `κʰ > 0`, the quantity this PR's whole argument rests on. Qwen2.5 is the paper's §5 limitation case and still shows it.

### The calibrated filter predicts true attention (Fig. 4)

Spearman vs **true** attention maps on held-out text:

| Scorer | Llama-3.2-1B | Llama-3.2-3B |
|---|---|---|
| **query-SVD (calibrated)** | **+0.783** (100% sign-correct) | **+0.863** (100%) |
| K-norm | +0.460 (94.5%) | +0.410 (94.6%) |
| key-SVD (fallback) | **−0.032** (46.1%) | **−0.008** (49.1%) |

Calibrated beats K-norm and reproduces the paper's Figure 4 ordering. The key-SVD fallback is **indistinguishable from noise** — sign correct less than half the time, worse than a coin flip. That is much stronger than the "sign is ambiguous" framing the original code used.

### ⚠️ End-to-end perplexity is blocked — no numbers claimed

Q-Filters doesn't remap RoPE positions after eviction. On a real model that means `cache.offset` tracks *retained* tokens, not true position — **128 after 300 tokens** at budget=128. `mlx_lm` derives query RoPE position from `cache.offset`, so post-eviction queries rotate at the wrong position and attention is scrambled. Perplexity explodes for **every** arm.

**Pre-existing on `master`** (verified by stashing this branch — same offset 128). Filed separately as #171. It degrades all policies equally, so no policy comparison is possible until fixed; the docs now say this explicitly instead of the old blanket "no model-level benchmark" caveat.

The correlation/anisotropy results are unaffected — they measure scoring quality directly, no generation loop involved.

Scripts committed for reproducibility; raw values in `figures/qfilters/real_model_results.json`. Full suite still 2198 passed.

rajveer43:
## Generation perplexity measured — #171 fixed (commit cf9a255)

The RoPE blocker is fixed, so the Figure-5-style comparison now runs.

### The fix

`mlx_lm` rotates **both** query and incoming key at `offset=cache.offset` *before* `update_and_fetch`. The base class left `offset` equal to the retained-row count, so it stalled at `budget` once eviction began — drift hit **+135 by token 199** at budget 64.

Reporting the true absolute position is sufficient: Q-Filters **preserves** original positions (drops rows, never renumbers), and RoPE is relative, so queries/new keys/survivors land on one consistent axis with **no re-rotation of survivors**. That's precisely why H2O/Keyformer need a delta-rotation pass and this cache doesn't. Calibrated ppl went **598.5 → 17.6**.

### Results (1024 tokens, non-repetitive prose)

**Llama-3.2-1B** — fp16 **4.050**

| Budget | Calibrated | Fallback | Gap closed |
|---|---|---|---|
| 256 (~4×) | **8.476** | 13.358 | **52%** |
| 128 (~8×) | **16.307** | 23.645 | **37%** |
| 64 (~16×) | **25.933** | 31.274 | **20%** |

**Llama-3.2-3B** — fp16 **3.305**

| Budget | Calibrated | Fallback | Gap closed |
|---|---|---|---|
| 256 (~4×) | **5.076** | 7.264 | **55%** |
| 128 (~8×) | **10.046** | 14.909 | **42%** |

Calibrated wins at **every budget on both models**, consistent with the attention-correlation ordering (+0.783/+0.863 vs −0.032/−0.008).

### Separate finding: `qfilters_recent` is effectively required for generation

Pure projection ranking is a *long-range importance* signal — it evicts the immediately-preceding tokens that next-token prediction leans on hardest. Llama-3.2-1B, budget 128, calibrated:

| `recent` | 0 | 32 | 64 | 96 |
|---|---|---|---|---|
| ppl | **263.2** | **16.3** | 20.3 | 24.9 |

A **16× swing**. This doesn't contradict the paper — it evaluates retrieval/NIAH, where recency matters far less. Left off by default to keep the default paper-faithful, now documented with a ≈budget/4 recommendation.

### Honest scope

These are a **policy comparison, not a reproduction of Figure 5** — perplexity is still well above fp16 at these ratios (uniform per-head budgets, no position renumbering, a much smaller calibration set than §4.2).

Two earlier harness drafts gave misleading numbers and were discarded: a single-pass prefill that evicts before scoring, and a repeated-paragraph corpus with fp16 ppl ~1.3 and no dynamic range. Both pitfalls are documented in the script.

**L2Norm is excluded** from the tables — it still carries the same un-fixed `offset` defect (`knorm_cache.py:165`) and would compare unfairly. Generalising the fix to the other evicting caches is worth a follow-up.

Still unmeasured: TTFT/throughput (Fig. 10), Ruler, NIAH, SnapKV/Expected Attention/StreamingLLM.

Tests: +2 RoPE regressions. Full suite **2200 passed**.

rajveer43:
## Follow-up: TTFT / throughput / NIAH, and a RoPE fix for the baselines

Three commits added since the last review, filling part of the **"Not measured"** list in the Q-Filters docs. Summary of what is measured, what was fixed to make measurement meaningful, and what is still explicitly out of scope.

| Commit | What |
|---|---|
| `e4186f7` | `fix(cache)`: RoPE offset after eviction in SnapKV + StreamingLLM |
| `2d93a41` | `docs(changelog)`: record that fix |
| `3e1377f` | `bench(qfilters)`: TTFT / throughput / NIAH harness |

---

### 1. The baselines were broken before they could be benchmarked

Both SnapKV-adapted and StreamingLLM-adapted carried the **same defect #171 fixed for Q-Filters**: `self.offset` reported the retained **row count** rather than the true absolute token position. `mlx_lm` rotates both the query and the incoming key at `offset=cache.offset` *before* `update_and_fetch` runs, so once eviction began every subsequent token was rotated at the wrong position.

Measured over 200 decode steps after a 64-token prefill at budget 32:

| Method | Offset reported | True position | Drift |
|---|---|---|---|
| **StreamingLLM** | frozen at **32** | 263 | **+231**, unbounded once the window saturates |
| **SnapKV** | 231 | 263 | **−32**, constant — exactly the tokens dropped during prefill compression |

**This had to be fixed first.** Benchmarking against the broken baselines would have shown Q-Filters winning by a wide margin, with the entire margin an artifact of position drift in its competitors. That is the same reasoning that kept L2Norm out of the perplexity table on this branch.

Both methods **preserve** original positions — `snap_select_indices` returns kept indices sorted ascending, and StreamingLLM's window drops rows without renumbering — and RoPE is relative, so reporting the true position is sufficient and survivors need no re-rotation.

#### SnapKV needed more than a counter

StreamingLLM took the same `_true_offset` shape as Q-Filters. SnapKV did not: its decode path appends **deltas**, so the base class's cursor arithmetic and return slice (`self.keys[..., :self.offset, :]`) genuinely need the row count. Overwriting `offset` broke three tests (`68 != 20` — the base class returned the buffer size instead of retained rows).

`offset` is therefore now a **property** yielding the row count while the base class is on the stack and the true position outside it, with `_process_prefill_chunk` reading `_row_offset` directly.

> **Reviewer note — a deliberate divergence from the paper.** StreamingLLM's paper assigns positions by index *within the cache*; we preserve original positions instead. Matching the paper would require re-rotating every survivor each step. This is recorded in the cache docstring, but it is a real behavioural difference and worth a second opinion.

Two existing tests asserted `offset == retained rows` — the old meaning — and were updated. A dedicated regression test was added that **fails without the fix**. Full suite: **2201 passed**.

---

### 2. TTFT and throughput

2048-token prompt, 64 decode steps, budget 256.

| Model | Method | TTFT | Decode tok/s |
|---|---|---|---|
| **Llama-3.2-1B** | fp16 | 1.876s | 102.5 |
| | Q-Filters | **1.819s** | 80.4 |
| | SnapKV | 2.276s | **111.0** |
| | StreamingLLM | 2.369s | 92.1 |
| **Llama-3.2-3B** | fp16 | **8.404s** | 39.3 |
| | Q-Filters | 9.210s | 32.8 |
| | SnapKV | 9.705s | **45.3** |
| | StreamingLLM | 9.524s | 37.1 |

SnapKV's decode advantage over fp16 widens with model size (+8% at 1B, **+15%** at 3B) as the smaller cache starts to pay for itself.

**These are not speedup claims for the algorithms.** All three caches run a per-`(B, H)` Python loop in `update_and_fetch`, so absolute tok/s reflects this implementation, not the methods' potential. Cross-method comparison at matched budget is the meaningful reading.

---

### 3. NIAH — and a Q-Filters defect it surfaced

fp16 scores **100% (18/18)** on both models, establishing the ceiling, so a lower score is a genuine compression effect rather than a malformed prompt.

At budget 256 (Llama-3.2-1B), every eviction method loses the needle — but **for three different reasons**, which is the part worth reviewing:

| Method | Acc. | Actual output |
|---|---|---|
| fp16 | 100% | `"The access code for the Aurora project is 7251."` |
| StreamingLLM | **33%** | `"...is not provided in the text."` — correct refusal |
| SnapKV | 0% | `"...is 1234."` — **fluent hallucination** |
| Q-Filters | 0% | `" the  the  the  the"` — **degenerate** |

StreamingLLM's 33% is mechanistically clean: every success is at depth **0.9**, i.e. the needle landed inside the trailing window. That consistency is good evidence the harness measures real behaviour.

#### Q-Filters is not losing the needle — it is emitting broken text

Investigating that `" the the the"` (initially suspecting my own harness) turned up a genuine issue. Output quality depends on **prefill chunk size**, at identical budget and identical 256 retained tokens:

| Prefill | Output |
|---|---|
| one-shot (1073 tok) | `" the  the  the  the"` — degenerate |
| 256-token chunks | partially coherent |
| ≤64-token chunks | `"...is not explicitly stated"` — coherent |

Cause is in [`qfilters.py`](../blob/master/veloxquant_mlx/quantizers/qfilters.py#L195) — `qfilters_update` *"absorbs a whole block of S tokens, then evicts down to budget **in one shot**"*, against a filter frozen on that same block. A single 1073→256 decision retains far less usable context than ~17 successive smaller ones.

**This affects the calibrated path too**, not just the key-SVD fallback — I verified both produce the same degenerate output one-shot. So it is *not* the sign ambiguity the docstring attributes to the fallback; the docstring's "path-DEPENDENT" wording understates the practical cost.

Since this harness prefills in one shot (what `model(prompt, cache=...)` does by default), **the Q-Filters rows above are its worst case.** I documented that in the script rather than quietly picking a chunk size that flatters the method.

> **Open question for review:** should `qfilters_update` evict incrementally rather than one-shot per block? That is a behavioural change to the method, so I have not made it here.

---

### Still not measured

Unchanged from the branch's existing disclosure, with the scope narrowed honestly:

- **RULER** beyond the NIAH family — the other ten task categories (variable tracking, common/frequent-word extraction, QA) are not covered. The retrieval arm here is NIAH only.
- **Expected Attention** — not implemented in this repo, so it cannot be a comparison arm. Adding it is a new method implementation, not a benchmark run.
- **L2Norm** — still carries the same un-fixed `offset` defect (`knorm_cache.py:165`) and would compare unfairly. Generalising this fix to it (and likely H2O/TOVA) is the obvious follow-up.

Qwen2.5-7B and a budget sweep (256/512/1024) are still running; I will post those numbers when they land rather than extrapolate from the two smaller models.

rajveer43:
### CI fix: docs build was broken since `570b7e2`

Two failures on this branch, both now green (`8669596`).

**1. Netlify deploy preview — broken since the blog post landed.** Docusaurus parses blog posts as **MDX**, where `<!-- truncate -->` is read as JSX and fails compilation:

```
Unexpected character `!` (U+0021) before name
  docs-site/blog/2026-08-14-qfilters-query-geometry.md line 24
```

Fixed to `{/* truncate */}` — the MDX form, which Docusaurus names as equally valid in its own untruncated-post warning. This is the only blog post in the repo using a truncation marker, which is why it had not surfaced before.

Netlify was `neutral` on the first branch commit and `failure` from `570b7e2` onward, so this predates the benchmark work. Verified with a local `npm run build` (now succeeds). The `blogs/` source carries no marker — it is mirror-only per `blogs/README.md` — so source and mirror remain in sync at 224/224 lines.

**2. commitlint** rejected `bench(qfilters):`; `bench` is not in the allowed type list. Amended to `test(qfilters):`.

All checks now pass, including the Netlify deploy preview.

Issue #277, opened 2026-08-30
research: Metal kernels for prefill-side attention compute (not just cache-read/decode)


rajveer43:
## Problem

Every Metal kernel in this repo today ([veloxquant_mlx/metal/](veloxquant_mlx/metal/)) accelerates something that happens to an **already-built** KV cache: `quantize_vq`, `rabitq_encode`, `rabitq_fused_attend`, `rabitq_pack_values`, `scalar_fused_decode_attend`. That's the right target for decode — decode is one query token against a long cache, so it's bandwidth-bound, and Apple Silicon's unified memory bandwidth (e.g. ~1.2 TB/s on M3 Ultra) is genuinely competitive there.

Prefill is a different regime. Prefill runs the full prompt through attention in one shot — S_q ≈ S_kv, both potentially in the tens of thousands to ~1M tokens for coding/agentic workloads where the whole codebase gets re-fed on every turn. That's compute-bound (large batched matmuls), and it's where Apple Silicon GPUs currently lose badly to CUDA hardware: back-of-envelope, an M3 Ultra running a large MoE model (e.g. GLM-class) sees on the order of ~300 tok/s prefill, versus ~10,000 tok/s on something like an RTX Pro 6000 — roughly a 30x gap. At 1M context that's the difference between a ~1M/300s (~55 min) wait before the first output token and a ~1M/10000s (~100s) wait. For coding agents that repeatedly re-submit a large, mostly-unchanged context, this makes Mac effectively unusable at long context lengths regardless of how good decode throughput is.

None of VeloxQuant-MLX's existing kernels touch this. Every one of them assumes the cache already exists and optimizes what happens when you read or write it afterward. The compression methods reduce cache *size*, and [cross-model KV transfer](https://veloxquant-mlx.netlify.app/docs/algorithms/cross-model-transfer) lets you skip prefill entirely between same-family models — but neither one makes a from-scratch prefill itself faster. There is currently no kernel work in this repo aimed at the prefill/attention compute path itself.

## Proposal

Investigate and prototype Metal kernels (via `mx.fast.metal_kernel`, matching the approach already used for `quantize_vq` and the RaBitQ/scalar fused-attend kernels) targeting the **prefill-side** attention compute, not just post-hoc cache access:

[...]

PR #280, opened 2026-08-30
Prefill-side attention: causal masking + from-scratch flash kernel investigation (#277)

rajveer43:
## Summary

Addresses #277 (research: Metal kernels for prefill-side attention compute).

- **`rabitq_prefill_attend(causal=True)`** — extends the existing VLM-cross-attention prefill kernel with causal self-attention support, using the same tail-alignment convention as `fused_sdpa.metal` (`q_abs = (S_kv - S_q) + q_pos`). Fixes a `NaN` edge case in the online-softmax update that causal masking exposes (fully-masked KV chunks).
- **`flash_prefill_attend`** (new) — a from-scratch, plain-fp16, causal-only `simdgroup_matrix` flash-attention kernel, purpose-built to try to beat `mx.fast.scaled_dot_product_attention` on the from-scratch prefill path (no pre-existing compressed cache to exploit). Specialized at compile time for exactly this repo's target shape: no GQA/mask/sinks branching, `exp2` softmax with pre-folded `log2(e)` scale, and a causal block-skip that drops fully-future KV chunks before any load/matmul work.
- **`scripts/prefill_roofline_bench.py`** (new) — self-calibrating roofline benchmark. Rather than trusting an unverified spec-sheet FLOPs peak, it measures achieved fp16 matmul throughput on the same GPU/run and compares SDPA against that.
- **`blogs/prefill-roofline.md`** (+ docs-site mirror) — full writeup of the investigation and results.

## Findings

Both prefill-shaped kernels are correctness-verified but **do not beat MLX's SDPA**:

- `rabitq_prefill_attend(causal=True)`: ~12-15x slower — the VLM-tuned decode-once-per-chunk trick doesn't amortize when `S_q == S_kv`.
- `flash_prefill_attend`: ~4-6x slower after four rounds of structural tuning (`BK` size, `exp2`, causal block-skip, SIMD-group-count sweep). `mx.fast.scaled_dot_product_attention` dispatches to MLX's own hand-tuned, AOT-compiled `steel` attention kernel — not a naive baseline — and the roofline shows it already captures most of this GPU's achievable fp16 matmul throughput (~85% in one run; see the blog for run-to-run variance caveats).

Net conclusion: the Mac-vs-CUDA prefill gap described in #277 looks like a hardware FLOPs ceiling, not a software gap this repo can close with a hand-written kernel. Full writeup with all numbers and what was tried is in `blogs/prefill-roofline.md`.

## Test plan

- [x] `python -m pytest veloxquant_mlx/tests/metal/` — 453 passed, 24 skipped (0 regressions)
- [x] `test_rabitq_prefill.py` — 88 passed, 12 skipped (causal parity vs numpy reference)
- [x] `test_flash_prefill.py` (new) — 40 passed, 12 skipped, including a targeted check that perturbing a future/masked key doesn't change earlier output rows
- [x] `python scripts/prefill_roofline_bench.py` and `python scripts/metal_rabitq_prefill_bench.py` run end-to-end and produce the numbers quoted in the blog post


rajveer43:
## Second tuning pass on `flash_prefill_attend`

Follow-up to the from-scratch kernel investigation above — a measurement-driven pass at closing the gap to SDPA, following an audit-then-measure discipline throughout (every change validated for correctness before being kept, kept only if a head-to-head benchmark showed a real win).

### What was tried

1. **Synchronization audit** — traced every `threadgroup_barrier`/`simdgroup_barrier` to its producer/consumer before touching anything. Result: every `threadgroup_barrier` was guarding genuinely cross-SIMD-group traffic (the shared `kv_tile` staging buffer); none could be safely removed.
2. **Register-resident online-softmax state** — `m_run`/`d_run`/`f_row` moved from threadgroup arrays to per-lane registers, broadcast via `simd_shuffle` only where cross-lane visibility is actually needed. Small, consistent win (+1-3%) at the largest shapes.
3. **W·V depth-tile batch (`PDT`) and KV chunk width (`BK`)**, generalized to template constants and swept per head-dimension rather than picked by a memory-budget heuristic. The obvious heuristic ("largest that fits 32KB") picked a `PDT` value that was **24-27% slower** at D=64 than the actual measured optimum — crossing the ~16KB double-occupancy line cost more than the barrier savings it bought.
4. **`BQ=16`** (doubling rows per SIMD-group) — implemented, fully validated for correctness, but measured **11-42% slower everywhere tested**. Kept as an unused template option (`BQ_ROWS`), not shipped.

### Results (Apple M4, 10-core GPU)

| D | S | flash (ms) | sdpa (ms) | sdpa/flash | vs. first pass |
|---|---|---|---|---|---|
| 32 | 512 | 1.29 | 1.53 | 1.19x | **flash now wins** |
| 32 | 2048 | 2.42 | 5.89 | **2.44x** | **flash now wins by >2x** (was competitive-but-behind) |
| 64 | 2048 | 6.61 | 1.89 | 0.29x | unchanged |
| 64 | 8192 | 88.0 | 26.5 | 0.30x | unchanged |
| 128 | 8192 | 234.3 | 52.9 | 0.23x | was 0.15x — gap shrank from ~6.7x to ~4.3x |

D=32 now genuinely beats SDPA. D=128's largest tested shape closed materially. D=64 is unchanged after three independent, correctly-measured attempts — its occupancy ceiling looks structural (already past the double-occupancy line at baseline, no room to grow into a wider tile without losing single-occupancy too), not something further tile-shape tuning fixes.

### Verification
- 90/90 correctness checks pass (new harness: `scripts/flash_prefill_harness.py`, covering first/last token, partial KV blocks, non-32-aligned query blocks, cache continuation)
- 128/128 existing pytest cases pass, 24 correctly skipped
- `ruff format --check .` clean (665 files)
- docs-site build succeeds

Full writeup with per-optimization detail: [`blogs/prefill-roofline.md`](https://github.com/rajveer43/VeloxQuant-MLX/blob/feat/prefill-causal-attention-277/blogs/prefill-roofline.md#step-5-a-second-pass-at-flash_prefill_attend--measure-everything-keep-only-what-wins) (Step 5).

Issue #435, opened 2026-09-18
Add documented per-file-ignores for B023 (scripts/) and E402 (benchmark_scripts/) — 167 verified false positives


rajveer43:
## Summary

167 of the repo's 529 ruff violations are **verified false positives** for
this codebase's conventions. They should be silenced with documented
`per-file-ignores`, not "fixed" — changing the code would make it worse in
both cases. Doing this is a prerequisite for making `ruff check` enforcing
(see the companion issue on blocking lint).

---

## Part 1 — B023 (137 hits, `scripts/`): all false positives

`ruff check . --select B023` reports 137 `function-uses-loop-variable`
across 8 benchmark harnesses:

```
 35 scripts/kv_kernel_gqa_packing_recheck.py
 23 scripts/metal_rabitq_prefill_bench.py
 23 scripts/kv_kernel_roofline_bench.py
 21 scripts/metal_rabitq_attend_bench.py
 14 scripts/prefill_roofline_bench.py
  8 scripts/flash_prefill_harness.py
  7 scripts/metal_rabitq_encode_bench.py
  6 scripts/kv_bookkeeping_microbench.py
```

B023 catches a real bug class — a closure capturing a loop variable reads
that variable's *final* value if it is called after the loop moves on. That
would silently corrupt benchmark numbers, so this was investigated as a
correctness question, not a style one.

**It does not happen here.** Every one of the 8 files defines its own timing
helper (`_bench`, `_timeit`, `bench_decode_step`) and every flagged closure
is passed straight into it. Each helper invokes `fn()` **synchronously inside
its own loops** and never stores, schedules, or returns the callable — e.g.
`scripts/kv_kernel_roofline_bench.py:60-65`:

```python
def _bench(fn, n_warmup, n_iter):
    for _ in range(n_warmup):
        mx.eval(fn())
    t0 = time.perf_counter()
    for _ in range(n_iter):
        mx.eval(fn())
    return (time.perf_counter() - t0) / n_iter
```

The closure is fully consumed before the enclosing loop rebinds anything, so
it always observes the intended iteration's values.

Checked mechanically for any deferral primitive across all 8 files:

```
$ grep -nE 'functools\.partial|\.append\(\s*(lambda|run|fn|step)\b|Thread|Pool|submit|asyncio|\.map\(|callbacks?\s*=|schedule' <the 8 files>
NONE FOUND
```

And confirmed the two semantics differ exactly as expected, using the repo's
own call pattern:

```
[...]

PR #443, opened 2026-09-18
docs: correct stale 3.11 Python-floor references left by #432 (#439)

### LINKED ISSUE #439: Fix stale "3.11 floor" reference in the [tool.mypy] comment (floor is 3.10 since #432)
## What was measured

The `[tool.mypy]` comment in `pyproject.toml:105-111` justifies its
`python_version` setting by contrasting it with the `requires-python` floor —
but names the wrong floor:

```toml
# 3.12, not the 3.11 floor in `requires-python`: numpy's bundled stubs use the
# PEP 695 `type X = ...` alias syntax, which mypy only parses when the
# configured target is 3.12+. Pinning this to 3.11 makes mypy fail outright
# ("Type statement is only supported in Python 3.12 and greater") against any
# numpy satisfying this project's own `numpy>=1.26` dependency floor, before
# it even reaches real code.
python_version = "3.12"
```

The actual floor is **3.10**, not 3.11:

```
$ grep -n 'requires-python\|python_version' pyproject.toml
10:requires-python = ">=3.10"
105:# 3.12, not the 3.11 floor in `requires-python`: numpy's bundled stubs use the
111:python_version = "3.12"
```

## Cause

26ec112 (`chore: lower Python floor to 3.10, matching mlx's own requirement
(#432)`) changed `requires-python = ">=3.11"` to `">=3.10"` and added the
`Programming Language :: Python :: 3.10` classifier, but did not update the
`[tool.mypy]` comment that refers back to that value:

```
$ git show 26ec112 -- pyproject.toml | grep -E '^[+-].*(requires-python|3\.1[012])'
-requires-python = ">=3.11"
+requires-python = ">=3.10"
+    "Programming Language :: Python :: 3.10",
```

## Why it matters

The reasoning in the comment is **correct and worth keeping** — the divergence
between the mypy target and the packaging floor is deliberate and non-obvious,
and without the explanation someone would "fix" it and break mypy. The problem
is only that it now misstates the value it is contrasting against, which makes
a reader check `requires-python`, find 3.10, and distrust the whole comment.

Low severity, documentation-only, but it sits in the one file a JOSS/PyPI
reviewer reads most closely.

## Scope

**Packaging metadata / tooling config.** No runtime or shipped-code effect.

## Fix

Update the two stale references in the comment (3.11 -> 3.10), keeping the
substance intact:

```toml
# 3.12, not the 3.10 floor in `requires-python`: numpy's bundled stubs use the
# PEP 695 `type X = ...` alias syntax, which mypy only parses when the
# configured target is 3.12+. Pinning this to 3.10 makes mypy fail outright
# ("Type statement is only supported in Python 3.12 and greater") against any
# numpy satisfying this project's own `numpy>=1.26` dependency floor, before
# it even reaches real code.
python_version = "3.12"
```

## Note for whoever picks this up

Everything else in the packaging metadata was audited at the same time and is
**correct** — no action needed there:

- `twine check` PASSES on a freshly built 0.88.1 wheel.
- `py.typed` ships (PEP 561), as do all 44 `.metal` kernel sources and
  `ui/static/*` — the `package-data` globs resolve correctly.
- `.zenodo.json`, `paper/joss/paper.md` and `pyproject.toml` agree on author
  name, ORCID (0009-0009-7566-1843), affiliation and MIT license.
- The 3.10–3.14 classifier list matches `requires-python = ">=3.10"` and the
  CI test matrix.

(The one real packaging defect found in that audit is filed separately: the
wheel also ships 184 test files.)

rajveer43:
## Summary

26ec112 lowered `requires-python` from `">=3.11"` to `">=3.10"` but did not sweep the places that quote that value back. This fixes the three that are still wrong.

**`pyproject.toml` `[tool.mypy]`** — the comment justified `python_version = "3.12"` by contrasting it with "the 3.11 floor in `requires-python`". The floor is 3.10, so the comment misstated the very value it was contrasting against: a reader checks `requires-python`, finds 3.10, and stops trusting an explanation that is otherwise correct and load-bearing.

I re-verified the reasoning before rewording it, rather than just swapping the numbers. With the target set to the floor, mypy still aborts on numpy's own bundled stubs with the exact error the comment quotes:

```
$ mypy --python-version=3.10 --ignore-missing-imports probe.py
.venv/lib/python3.12/site-packages/numpy/__init__.pyi:737: error: Type statement is only
  supported in Python 3.12 and greater  [syntax]
Found 1 error in 1 file (errors prevented further checking)

$ mypy --python-version=3.12 --ignore-missing-imports probe.py
Success: no issues found in 1 source file
```

So the divergence stays, and the comment keeps explaining it. I also added a line stating that `python_version` is a type-checking target and *not* a support claim — that misreading is what made the stale number matter in the first place.

**Live user-facing docs** — the same sweep was missing from two surfaces that actively misinform a 3.10 user that they are unsupported, when `pip install` works for them:

- `docs-site/docs/getting-started/installation.md` requirements table: `3.11 or 3.12` → `3.10 – 3.14`
- `landing/index.html` requirement badge (`PYTHON 3.11–3.12` → `PYTHON 3.10+`) and FAQ answer. That FAQ sentence is duplicated in a `application/ld+json` structured-data block and in the visible copy; both were updated together so the structured data keeps matching the rendered page. All four JSON-LD blocks re-validated as parseable afterwards.

### Deliberately not changed

Dated blog posts under `blogs/` and `docs-site/blog/2026-05-*` also say 3.11. They were accurate when published, and the version line there sits inside the environment description of a benchmark run — retroactively editing it would falsify the record of what was actually executed. Left alone on purpose.

### 3.10 is a real floor

Not an aspiration: `mlx` itself declares `Requires-Python: >=3.10`, and both `mlx-tests.yml` and `non-metal-unit.yml` run the suite across `["3.10", "3.11", "3.12", "3.13", "3.14"]`.

## Related issue

Closes #439

## Test plan

- [x] `python -m pytest veloxquant_mlx/tests -q` passes locally on Apple Silicon — **3770 passed, 116 skipped**, matching the baseline
- [x] `pyproject.toml` still parses; `requires-python` and `python_version` read back as `>=3.10` / `3.12`
- [x] All 4 `application/ld+json` blocks in `landing/index.html` validate as JSON after the edit
- [x] `ruff check veloxquant_mlx` output is byte-identical before and after this branch (pre-existing findings tracked in #435/#436/#438 are unaffected)
- [ ] New or updated unit tests cover the change — n/a, no executable code changed; the diff is comments plus four documentation strings
- [x] Docs updated if user-facing behavior changed

## Number claims (if any)

None. No compression, speedup, memory or throughput claim is made or altered — the only numbers in this diff are Python version identifiers, verified against `requires-python`, mlx's own metadata, and the CI matrix.


rajveer43:
Added a follow-up commit (`c4186a4`) with the drift check discussed on #440 — so this class of bug fails CI instead of surviving until someone reads the file.

**Where it lives:** `veloxquant_mlx/tests/core/test_python_floor_consistency.py`, not a new workflow. It runs inside the existing test job, so it executes on all ten matrix legs (3.10–3.14 × `mlx-tests` + `non-metal-unit`) at no extra CI cost and with no new workflow to maintain. This follows the existing `sync_release_badges.py` / `test_sync_release_badges.py` pattern.

**What it asserts** (8 checks): every stated floor in README, the docs-site install table, and the landing badge + FAQ equals `requires-python`; the landing FAQ's JSON-LD and visible copies agree with each other; the trove classifiers start at the floor and have no gaps; and the `[tool.mypy]` comment does not misname the floor it contrasts against.

**Verified by mutation, not by passing.** A green consistency test proves nothing until it's shown to fail on the drift it targets:

| Mutation | Result |
|---|---|
| Revert floor to `>=3.11`, docs untouched (the literal #432 scenario) | **7 of 8 fail** |
| Desync landing FAQ JSON-LD from visible copy | **sync check fails alone** |
| Stale only the docs-site install table | **that row fails alone** |

**Verified against false positives**, since a noisy guard gets deleted:

| Legitimate edit | Result |
|---|---|
| Bump `brew install python@3.12` / `conda create ... python=3.12` examples | **stays green** — naming a known-good version to install isn't a claim about the supported range |
| `[tool.mypy] python_version = "3.12"` staying above the floor | **stays green** — the check is that the comment doesn't *misname* the floor, not that it equals it |

Rewording a doc past its pattern fails loudly (`update the pattern in _FLOOR_CLAIMS so this stays a real check instead of silently passing`) rather than quietly matching nothing — that silent-no-op is exactly how `sync_release_badges` drifted for several releases before being caught.

**One real bug caught while writing it:** the first draft used `tomllib`, which is stdlib only from 3.11. A bare import would have `ImportError`'d at *collection* on the 3.10 leg — the precise oldest-Python job this test exists to protect — taking the whole run down with it. Now parsed with `re`; adding a `tomli` backport to read two scalar fields would have meant taking a dependency in order to defend a floor. Confirmed the module imports and both helpers work with `tomllib` blocked.

Suite: **3778 passed, 116 skipped** (3770 + 8 new, no regressions).

